PR #17, opened 2020-04-17
Personalize delete custom resource

### LINKED ISSUE #4: Custom resource to delete Personalize resources
Since the Personalize resources (schemas, datasets, solutions, campaigns, event tracker) are all created outside of CloudFormation by a [notebook](https://github.com/aws-samples/retail-demo-store/blob/master/workshop/1-Personalization/1.1-Personalize.ipynb) or Lambda function, they are not cleaned up automatically when the Retail Demo Store stack is deleted. Although there is a [Personalize Cleanup Notebook](https://github.com/aws-samples/retail-demo-store/blob/master/workshop/1-Personalization/1.2-Personalize-Cleanup.ipynb) that can be used to delete these resources, it would be easy to miss or forget to run through this notebook before deleting the stack. Once the stack is deleted, the notebooks are deleted too. In addition, if the user initially deploying the stack chooses to have the Personalize campaigns auto-created post-deployment, it's not apparent that the cleanup notebook should be run. This could leave behind resources that will continue to incur unexpected charges. 

To ensure that Personalize resources created as part of the Retail Demo Store deployment are cleaned up when the stack is deleted, a custom resource should be created that deletes these resources.

james-jory:
*Issue #, if available:*

Closes #4 

*Description of changes:*

- Created new lambda function that deletes all Personalize resources created by the project. 
- Created custom resource that calls the new lambda function when the stack is deleted. This will ensure that Personalize resources are not left behind leading to unexpected AWS expenses when the user deletes the CloudFormation stack.
- Re-organized some of the deployment related support resources into separate templates, moving them out of the root template file to keep cleaner. This will also help when creating and updating modules and blueprints in Event Engine for group workshops (i.e. just the defaults in a streamlined root template need to be set).


aws/cloudformation-templates/cleanup-bucket.yaml
```diff
@@ -0,0 +1,85 @@
...
+                Resource: '*'
+        - PolicyName: S3Policy
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action:
+                  - s3:List*
+                  - s3:DeleteObject
+                Resource: '*'
```
BastLeblanc, line 80: Can we scope down to the bucket used?

PR #22, opened 2020-04-29
Chatbot workshop

### LINKED ISSUE #10: Build conversational AI workshop notebook
The current Lex bot integration on the Help view (accessed from "?" icon in top nav) will load the chat bot widget if there is a Lex bot configured on the backend. Otherwise, a message is displayed indicating that it's not fully configured. 

Building out the chatbot experience would make for a useful workshop. For this issue, a workshop notebook should be created that walks the user/developer through building a chatbot in Lex that provides retail-specific intents. Once the bot is created on the backend via the workshop, the frontend is already setup to automatically load and display the bot widget.

james-jory:
*Issue #, if available:*

Closes #10 

*Description of changes:*

I needed a conversational/chatbot workshop to round some features for an upcoming blog post.

- Required working around limitations in the Amplify chatbot component where session attributes could not be sent with PostText/PostContent calls (i.e. to pass the user ID) and the component does not support response cards. Linking the entityId sent by the chatbot to a store user required tracking the entityId with the user in the Users service. Unfortunate this was necessary but a workaround was needed. This was the cleanest approach that I could get working.
- A new Lambda function was added that is called by Lex to retrieve recommendations.
- A new workshop notebook was created to walk reader through setting up the chatbot and wiring up the Lambda function.


src/aws-lambda/bot-intent-recommendations/bundle.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
BastLeblanc, line 1: general comment on the bundle.sh and stage.sh.
It's the same for every Lambda function, maybe some opportunities to factor some code?

.gitignore
```diff
@@ -16,3 +16,4 @@ demo.md
 generators/*.gz
 /csvs/
 .unotes/*
+!workshop/5-Conversational/RetailDemoStore_Lex.zip
```
BastLeblanc, line 19: why this zip in the gitignore? need it for import in Lex console, right?

src/aws-lambda/bot-intent-recommendations/bundle.sh
james-jory: Yes, there is a growing need to simplify how the growing number of lambdas are deployed. SAM is probably the answer.

.gitignore
james-jory: Yes, the zip file is needed to import the chatbot config in Lex. It's very small. Just contains a JSON file.

Issue #21, opened 2020-04-28
Improve CI/CD automation


BastLeblanc:
After each commit on master, each regional s3 bucket should be updated.

- run stage.sh for each bucket, currently us-west-2, us-east-1 and eu-west-1
- notify of the deployment / alerts in case of failure to the team.

A live version of the demo could be also updated at the same time.

BastLeblanc:
Closing this issue, internal account ready and #25 merged

PR #25, opened 2020-05-06
codebuild automation

BastLeblanc:
*Issue #, if available:* #21 

*Description of changes:*
Automate deployment of commit to buckets

- added buildspec.yml
- added --acl public-read for the stage script to publish objects on public in s3
- adapted the README.md to the new buckets retail-demo-store-XX-region-X


buildspec.yml
```diff
@@ -0,0 +1,23 @@
+version: 0.2
```
james-jory, line 1: Add a comment describing how/where this file is used so as not to confuse repo users.

stage.sh
```diff
@@ -35,30 +38,30 @@ find . -name '.DS_Store' -type f -delete
...
+aws s3 cp src/products/src/products-service/data/ s3://${BUCKET}/${S3PATH}data --recursive  $S3PUBLIC
+aws s3 cp src/users/src/users-service/data/ s3://${BUCKET}/${S3PATH}data --recursive $S3PUBLIC
 
 # Sync CSVs used for Personalize pre-create campaign Lambda function
 echo " + Copying CSVs for Personalize model pre-create training"
-aws s3 sync s3://theglobalstore-code/csvs s3://${BUCKET}/${S3PATH}csvs --only-show-errors
+aws s3 sync s3://theglobalstore-code/csvs s3://${BUCKET}/${S3PATH}csvs --only-show-errors $S3PUBLIC
 
 # Stage AWS Lambda functions
 echo " + Staging AWS Lambda functions"
```
james-jory, line 67: Looks like you're missing the new 'bot-intent-recommendations' lambda in the for loop. Is your branch up to date?

stage.sh
```diff
@@ -35,30 +38,30 @@ find . -name '.DS_Store' -type f -delete
...
 echo " + Upload seed data"
-aws s3 cp src/products/src/products-service/data/ s3://${BUCKET}/${S3PATH}data --recursive 
-aws s3 cp src/users/src/users-service/data/ s3://${BUCKET}/${S3PATH}data --recursive
+aws s3 cp src/products/src/products-service/data/ s3://${BUCKET}/${S3PATH}data --recursive  $S3PUBLIC
+aws s3 cp src/users/src/users-service/data/ s3://${BUCKET}/${S3PATH}data --recursive $S3PUBLIC
 
 # Sync CSVs used for Personalize pre-create campaign Lambda function
 echo " + Copying CSVs for Personalize model pre-create training"
-aws s3 sync s3://theglobalstore-code/csvs s3://${BUCKET}/${S3PATH}csvs --only-show-errors
+aws s3 sync s3://theglobalstore-code/csvs s3://${BUCKET}/${S3PATH}csvs --only-show-errors $S3PUBLIC
```
james-jory, line 64: The CSVs should be staged in a bucket in the deployment account rather than "theglobalstore-code" bucket. This should remove the final external dependency.

stage.sh
BastLeblanc: Thanks, added.

stage.sh
BastLeblanc: " in the deployment account "
Which one you refering to?

Do you think the csv should be commited in the repo ?

buildspec.yml
BastLeblanc: We can also not use a buildspec.yml to avoid confusion as it's not use for launching the demo in a customer account

james-jory: No, I think it's okay. Was just thinking a comment would help eliminate any potential confusion as to the purpose of the file.

stage.sh
james-jory: The new account where the new buildspec is being deployed. I was thinking that putting the CSVs in a bucket in that account is better than where they are now. I'd prefer to keep the CSVs out of the repo since the interactions file is quite large and the items file will get bigger as more products are added.

BastLeblanc:
changes:
- instructions on buildspec.yml
- moved csv to retail-demo-store-code bucket
- bot-intent lambda support

PR #43, opened 2020-06-15
Recipe parameter for recommendations events

manbearshark:
*Issue #, if available:*
Recipe needs to be available for analytics tools to detect that a product view was a result of a personalization feature in the user interface.

*Description of changes:*

Added a "recipe" parameter to the URL of products that are added to the UI as a result of a particular personalization recommendation strategy.  Currently only in the category detail rank feature but if this looks OK to reviewers will also add this to the other features.  This will help to measure the actual conversion on recommended products via tools like Amplitude.


src/web-ui/src/analytics/AnalyticsHandler.js
```diff
@@ -197,6 +197,7 @@ export const AnalyticsHandler = {
                 category: product.category,
                 image: product.image,
                 feature: feature,
+                recipe: recipe,
```
james-jory, line 200: Please add 'recipe' to attributes for AmplifyAnalytics.record call above too for consistency.

src/web-ui/src/analytics/AnalyticsHandler.js
```diff
@@ -319,6 +320,7 @@ export const AnalyticsHandler = {
...
```
james-jory, line 323: Same comment here. Add 'recipe' to AmplifyAnalytics.record call above.

src/web-ui/src/public/CategoryDetail.vue
```diff
@@ -108,11 +110,13 @@ export default {
...
           if (response.headers['x-personalize-recipe']) {
             this.personalized = true
-            this.explain_recommended = 'Personalize recipe: ' + response.headers['x-personalize-recipe']
+            this.recipe = response.headers['x-personalize-recipe']
+            this.explain_recommended = 'Personalize recipe: ' + this.recipe
           }
           if (response.headers['x-experiment-name']) {
             this.active_experiment = true
-            this.explain_recommended = 'Active experiment: ' + response.headers['x-experiment-name']
+            this.experiment = response.headers['x-experiment-name']
```
james-jory, line 118: Is `this.experiment` needed? Doesn't look like it's used anywhere besides the next line.

src/web-ui/src/public/Main.vue
```diff
@@ -181,11 +184,13 @@ export default {
...
         if (response.headers['x-personalize-recipe']) {
           this.personalized = true
-          this.explain_recommended = 'Personalize recipe: ' + response.headers['x-personalize-recipe']
+          this.recipe = response.headers['x-personalize-recipe']
+          this.explain_recommended = 'Personalize recipe: ' + this.recipe
         }
         if (response.headers['x-experiment-name']) {
           this.active_experiment = true
-          this.explain_recommended = 'Active experiment: ' + response.headers['x-experiment-name']
+          this.experiment = response.headers['x-experiment-name']
```
james-jory, line 192: Is `this.experiment` needed? Doesn't look like it's used anywhere besides the next line.

src/web-ui/src/public/CategoryDetail.vue
manbearshark: The idea was to have this available in the Amplitude A/B testing workshop which they wanted to add after the experimentation one.

james-jory: > The idea was to have this available in the Amplitude A/B testing workshop which they wanted to add after the experimentation one.

Okay, so another parameter passed to AnalyticsHandler.js calls? If so, why not weave it all the way through now so it's passed to Amplitude (and other event sinks)?

manbearshark: It looks like at some point the productViewed AnalyticsHandler function changed to take an experimentCorrelationId vs an experiment parameter.  Just want to make sure this will not be a breaking change as the only call I can see to this function is in the ProductDetail component.

Issue #38, opened 2020-06-04
add support for Stock management for products


BastLeblanc:
The idea is to be able to integrate with external demos which would simulate physical stores, and stock could be updated via an API.

Design is probably simply a new field in the product service and an api

Issue #9, opened 2020-04-14
Index data into ES using DDB stream + Lambda


james-jory:
Rather than indexing products directly in Elasticsearch using the Lambda pre-loader (custom resource) or the Search workshop notebook, use a DDB stream off the products table to index products in ES automatically. This will keep the ES index in sync with the source of truth in DDB as well as make it easier to manage once product editing is supported.

BastLeblanc:
I'm highly in favor of DDB streams and Lambda functions.

We should decouple problematic (it's a Retail Demo, but also a way to demonstrate architecture best practices)

PR #49, opened 2020-06-24
products API : update and inventory

BastLeblanc:
*Issue #, if available:* solving #38 

*Description of changes:*
Adding 2 new APIs in the product service:
	POST:	"/products/" : updates a product
	POST:     "/products/inventory" updates a product stock

Small UI tweak to add stock and disabled add to cart if stock = 0

To use the API   "/products/" to update a product, simply POST the full json from a product, for example:
```
{"id":"1",
"url":"http://xxxx.cloudfront.net/#/product/1",
"sk":"",
"name":"Black Leather Backpack",
"category":"accessories",
"style":"bag",
"description":"Our handmade leather backpack will look great at the office or out on the town.",
"price":119.99,
"image":"1.jpg",
"featured":"true",
"currentstock":5}
```

To use the API   "/products/inventory" to update a product stock, simply use a POST with a body containing product id and the stock delta (which can be negative or positive):

```
{
    "id": "1",
    "stockdelta": 10
}
```


src/products/src/products-service/repository.go
```diff
@@ -373,3 +373,75 @@ func RepoFindALLProduct() Products {
...
+		fmt.Println("Got error calling PutItem:")
+		fmt.Println(err.Error())
+
+	}
+
+	return product
+
+}
+
+func RepoUpdateInventory(inventory Inventory) Product {
```
james-jory, line 405: Would an AtomicCounter operation be more appropriate here to avoid concurrent update issues?

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithItems.html#WorkingWithItems.AtomicCounters

src/products/src/products-service/routes.go
```diff
@@ -53,4 +53,16 @@ var routes = Routes{
...
 	},
+	Route{
+		"ProductUpdate",
+		"POST",
+		"/products/",
+		UpdateProduct,
+	},
+	Route{
+		"InventoryUpdate",
+		"POST",
```
james-jory, line 64: Should this be a PUT rather than POST since we are updating an existing product?

src/products/src/products-service/routes.go
```diff
@@ -53,4 +53,16 @@ var routes = Routes{
...
+		"/products/inventory",
```
james-jory, line 65: Should the path be `/products/inventory/{productID}` so it's in the resource context of a product?

src/products/src/products-service/product.go
```diff
@@ -17,7 +17,14 @@ type Product struct {
 	Image          string  `json:"image" yaml:"image"`
 	Featured       string  `json:"featured,omitempty" yaml:"featured,omitempty"`
 	GenderAffinity string  `json:"gender_affinity,omitempty" yaml:"gender_affinity,omitempty"`
+	CurrentStock   int     `json:"currentstock" yaml:"currentstock"`
```
james-jory, line 20: JSON and YAML names should be `current_stock` for consistency with other multi-word attributes.

src/products/src/products-service/data/products.yaml
```diff
@@ -6,6 +6,7 @@
   price: 109.99
   image: "1.jpg"
   featured: true
+  currentstock: 1
```
james-jory, line 9: See my comment for product.go below about naming `current_stock` instead.

src/products/src/products-service/handlers.go
```diff
@@ -162,3 +164,65 @@ func ProductFeatured(w http.ResponseWriter, r *http.Request) {
 func enableCors(w *http.ResponseWriter) {
 	(*w).Header().Set("Access-Control-Allow-Origin", "*")
 }
+
+// Update Stock Quantity for one item Handler
```
james-jory, line 168: Change documentation to match function.

src/products/src/products-service/product.go
```diff
@@ -17,7 +17,14 @@ type Product struct {
...
 }
 
 // Products Array
 type Products []Product
+
+// Inventory Struct
+type Inventory struct {
+	ID         string `json:"id" yaml:"id"`
+	StockDelta int    `json:"stockdelta" yaml:"stockdelta"`
```
james-jory, line 29: JSON and YAML names should be `stock_delta` for consistency.

src/products/src/products-service/product.go
```diff
@@ -17,7 +17,14 @@ type Product struct {
...
 	GenderAffinity string  `json:"gender_affinity,omitempty" yaml:"gender_affinity,omitempty"`
+	CurrentStock   int     `json:"currentstock" yaml:"currentstock"`
 }
 
 // Products Array
 type Products []Product
+
+// Inventory Struct
+type Inventory struct {
+	ID         string `json:"id" yaml:"id"`
```
james-jory, line 28: ID may not be necessary if specified on path.

src/web-ui/src/public/ProductDetail.vue
```diff
@@ -214,6 +215,22 @@ export default {
...
+    },
+    productStock: function () {
+      if (this.product.currentstock > 0 ) {
+        return "Items currently in stock : " + this.product.currentstock 
+      }
+      else {
+        return "Sorry this item is currently out of stock"
+      }
+    },
+    addToCartisDisabled: function () {
```
james-jory, line 227: The lowercase "is" looks odd here. Either change to `addToCartIsDisabled` or `addToCartDisabled` so it's fully camel case.

src/web-ui/src/public/ProductDetail.vue
```diff
@@ -214,6 +215,22 @@ export default {
...
+      if (this.product.currentstock > 0 ) {
```
james-jory, line 228: Consider reducing function to single line `return !this.product.current_stock > 0` for brevity.

src/web-ui/src/public/ProductDetail.vue
```diff
@@ -214,6 +215,22 @@ export default {
...
         return root_url + this.product.category + '/' + this.product.image
       }
+    },
+    productStock: function () {
+      if (this.product.currentstock > 0 ) {
+        return "Items currently in stock : " + this.product.currentstock 
+      }
+      else {
+        return "Sorry this item is currently out of stock"
+      }
```
james-jory, line 225: The template is a better place for messages rather than in code. Keeps the presentation elements together. You can use Vue [conditional rendering](https://vuejs.org/v2/guide/conditional.html) to display different messages based on the stock level. So something like this:

```html
<span v-if="product.current_stock > 0">Items currently in stock: {{ product.current_stock }}</span>
<span v-else>Sorry, this item is currently out of stock</span>
```

src/products/src/products-service/handlers.go
```diff
@@ -162,3 +164,65 @@ func ProductFeatured(w http.ResponseWriter, r *http.Request) {
...
+		w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+		w.WriteHeader(422) // unprocessable entity
+		if err := json.NewEncoder(w).Encode(err); err != nil {
+			panic(err)
+		}
+	}
+
+	log.Println("UpdateProducts  ", product)
+
+	ret := RepoUpdateProduct(product)
```
james-jory, line 191: How will product add vs update work? If an `id` is specified, should an existing product be updated. If `id` is absent, add a new product?

src/products/src/products-service/data/products.yaml
james-jory: Should the stock levels be higher than 1, at least for some products?

src/products/src/products-service/repository.go
BastLeblanc: I guess you meant Conditional Writes ? https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithItems.html#WorkingWithItems.ConditionalUpdate

I will try to implement it

src/products/src/products-service/data/products.yaml
BastLeblanc: I've put random numbers between 1-15

src/products/src/products-service/handlers.go
BastLeblanc: I propose:

Update existing:

		"PUT",
		"/products/{id}",
		UpdateProduct,
---
create new

		"POST",
		"/products/",
		NewProduct,

---
update stock

		"PUT",
		"/products/inventory/{id}",
		UpdateInventory,

src/products/src/products-service/repository.go
james-jory: > I guess you meant Conditional Writes ? 

It depends on the semantics of the API. If the API is intended to be, "increase/decrease inventory by N", then an atomic counter seems appropriate. If the API is intended to be, "change/set the inventory to N", then a conditional write seems appropriate. 

I could see use-cases for both, actually. When completing an order during checkout, we're looking to decrement the inventory. When processing a return or voiding an order, we're looking to increment the inventory. The current value doesn't matter (except, I suppose, when trying to decrement when the inventory is zero).

When performing an inventory audit where you're counting inventory, we're looking to update/set the inventory value.

src/products/src/products-service/handlers.go
james-jory: Looks good to me.

Another thought on adding products, which doesn't necessarily need to be solved in this PR, is how product images will be added. Perhaps this is solved when the UI changes are introduced to add/update products where an image upload to S3 (and resize?) is supported.

src/products/src/products-service/routes.go
BastLeblanc: corrected to PUT

src/products/src/products-service/routes.go
BastLeblanc: corrected to 		"/products/inventory/{productID}",

src/products/src/products-service/product.go
BastLeblanc: correct, for now only StockDelta is supported, but we can imagine more fields in the future (like absolute value for stock, adding metadata like stores, etc.)

src/products/src/products-service/handlers.go
BastLeblanc: For images for now you need 2 steps when creating a new product via API:
- upload 1 image to the relevant folder on s3
- PUT json product data with reference to the image filename in the payload

I would see these 2 operations done from the UI perspective. 

For a later PR

BastLeblanc:
almost done with this PR, need to implement Atomic counters or Conditional Writes for stock updates

BastLeblanc:
@james-jory  please comment and check last commit including conditional update on inventory

james-jory (changes requested):
Important tie-in with #9 here so that product adds/updates/deletes in DDB are sync'd with dependent services like ES for product search and Personalize for filters and cold starting items.

Detailed comments also added in this review.

src/products/src/products-service/routes.go
```diff
@@ -53,4 +53,22 @@ var routes = Routes{
...
+	Route{
+		"ProductUpdate",
+		"POST",
+		"/products/",
+		NewProduct,
+	},
+	Route{
+		"InventoryUpdate",
+		"PUT",
+		"/products/inventory/{productID}",
```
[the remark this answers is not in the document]
james-jory: Changed path to:

`/products/id/{productID}/inventory`

src/products/src/products-service/repository.go
```diff
@@ -373,3 +374,114 @@ func RepoFindALLProduct() Products {
 
 	return f
 }
+
+func RepoUpdateProduct(id string, updated_product Product) Product {
+
+	var existing_product Product
+	existing_product = RepoFindProduct(id)
```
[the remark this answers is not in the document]
james-jory: Added logic to make sure product already exists and if not, return a 404.

src/products/src/products-service/product.go
```diff
@@ -17,7 +17,13 @@ type Product struct {
...
 	GenderAffinity string  `json:"gender_affinity,omitempty" yaml:"gender_affinity,omitempty"`
+	CurrentStock   int     `json:"current_stock" yaml:"current_stock"`
 }
 
 // Products Array
 type Products []Product
+
+// Inventory Struct
+type Inventory struct {
+	StockDelta int `json:"stockdelta" yaml:"stockdelta"`
```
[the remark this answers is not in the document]
james-jory: Done

james-jory:
Made several changes to make this service easier to develop with, refactored some inefficient code, and made the error handling more robust.

- You can now run DDB locally to make developing, debugging, and testing easier. This now the default setup in docker-compose. However, you can still connect to DDB in your AWS account (if needed/desired). When running DDB locally, the products and categories tables are conditionally created and loaded from the products and categories yaml files.
- Changed the way URLs were fully qualified to be fully qualified by default and to use pointers rather than passing products and categories by value to make this logic more efficient.
- Always make sure the Content-Type is set to application/json when returning a success response (was inconsistent before).
- Improved error handling to properly return 404s and 422s.
- Added validation logic for product info when adding/updating products (checks on name, stock, price, category, etc).
- Fixes for properly returning empty arrays when there is no data to return (was returning "null" before).
- Added a new API endpoint to get a category by its ID. I needed to be able to lookup categories for validation anyway so decided to add a new endpoint for this too.
- Made several code changes recommended by golang linter (variable names, function comments, etc).

james-jory (approved):
Sync'd with upstream and tested all changes. Good to go.

BastLeblanc:
Great stuff @james-jory thanks for the last effort making this good to go.
Merging now and let's make sure the other PR which are tied to products are aligned to this

PR #53, opened 2020-07-28
Filters cleanup, lodash, load all users

james-jory:
*Description of changes:*

- Delete Personalize filters created in workshop
- Upgrade lodash dependency in web-ui
- Test users increased from 5k to 6k. Update experiments to raise max users so all are loaded.

### DIFF
--- a/src/aws-lambda/personalize-delete-resources/bundle.sh
+++ b/src/aws-lambda/personalize-delete-resources/bundle.sh
@@ -6,6 +6,7 @@ PACKAGE_FILE=personalize-delete-resources.zip
 
 echo "Cleaning up intermediate files"
 [ -e ${PACKAGE_FILE} ] && rm ${PACKAGE_FILE}
+[ -e "models" ] && rm -rf models
 [ -e "package" ] && rm -rf package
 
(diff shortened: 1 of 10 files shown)


src/aws-lambda/personalize-delete-resources/bundle.sh
```diff
@@ -16,7 +17,12 @@ cd package
 zip -r9 ${OLDPWD}/${PACKAGE_FILE} .
 cd ${OLDPWD}
 
+echo "Downloading Personalize SDK model files"
+wget -q https://raw.githubusercontent.com/boto/botocore/e22aad94ba9f33bda7a0abea3a959fb1af56b25d/botocore/data/personalize-runtime/2018-05-22/service-2.json -P ./models/personalize-runtime/2018-05-22
```
BastLeblanc, line 21: Are we sure these URLS are permanent? Shouldn't it be something like
https://raw.githubusercontent.com/boto/botocore/master/botocore/data/personalize/2018-05-22/service-2.json

(with the right branch and version 2018-05-22 )  ?

or include the json in the repo if we want to fix a version?

PR #54, opened 2020-08-04
Add AWS_DEFAULT_REGION environment variable

yeeland:
boto3 seems to look for this for the default region according to [docs](https://github.com/boto/boto3/blob/959bc56ead8640feb665b828c0f548194a9c4eb7/docs/source/guide/configuration.rst#using-environment-variables)

### DIFF
--- a/src/docker-compose.yml
+++ b/src/docker-compose.yml
@@ -25,6 +25,7 @@ services:
     container_name: products
     environment:
       - AWS_REGION
+      - AWS_DEFAULT_REGION
       - AWS_ACCESS_KEY_ID
       - AWS_SECRET_ACCESS_KEY
       - AWS_SESSION_TOKEN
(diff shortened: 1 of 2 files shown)


src/docker-compose.yml
```diff
@@ -25,6 +25,7 @@ services:
     container_name: products
     environment:
       - AWS_REGION
+      - AWS_DEFAULT_REGION
```
james-jory, line 28: The products service is written in golang and the [AWS go SDK looks for AWS_REGION](https://github.com/aws/aws-sdk-go/issues/2103). Is AWS_DEFAULT_REGION needed for the recommendations service and not products service?

src/docker-compose.yml
```diff
@@ -44,6 +45,7 @@ services:
       - products
     environment:
       - AWS_PROFILE
```
james-jory, line 47: This should be deleted then?

PR #62, opened 2020-08-20
Change order of recommendations service Dockerfile steps

yeeland:
*Description of changes:*
This prevents running `apt-get update && apt-get install -y g++ git` and `pip install -r requirements.txt` unless there were a change to `requirements.txt`

### DIFF
--- a/src/recommendations/Dockerfile
+++ b/src/recommendations/Dockerfile
@@ -2,6 +2,7 @@ FROM python:3.8-slim
 
 WORKDIR /app
 
+RUN apt-get update && apt-get install -y g++
 COPY /src/recommendations-service/requirements.txt /app/requirements.txt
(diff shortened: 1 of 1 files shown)


src/recommendations/Dockerfile
```diff
@@ -1,12 +1,11 @@
 FROM python:3.8-slim
 
-COPY /src/recommendations-service /app
-
 WORKDIR /app
 
-# Needed for Optimizely SDK (git ref in req.txt)
-RUN apt-get update && apt-get install -y g++ git
```
james-jory, line 8: I was partially wrong on this one. Removing this line entirely doesn't work for me. The image build fails. We still need g++ for an Optimizely SDK dependency but don't need git. So this line works for me.

`RUN apt-get update && apt-get install -y g++`

Issue #52, opened 2020-07-23
Please name related created resources uniquely


deathweaselx86:
Someone at my org deployed this, but didn't tear it down. Late, a different person attempted to deploy this but it failed because `RetailDemoStore-users-TaskRole already exists`. I would expect all of these resources to be named uniquely. Can we make that happen?

james-jory:
Thanks, @deathweaselx86. Most resources are uniquely named during deployment but some have fixed names since they are accessed by name in code in the workshop notebooks. Making all resources uniquely named would require the notebooks to resolve these names. We need to find an elegant approach to this that does not add confusing complexity to the notebooks. You're right, though, that we need to figure this out. As for the TaskRole's, we should be able to use dynamic names for those.

james-jory:
We will work on this issue in phases where the goal of the first phase will be to support multiple deployments in a single account but just in different regions. Some research and testing needs to be done to verify but this should just involve making sure that IAM resources are uniquely named. This includes IAM roles in the CloudFormation templates and roles created in notebooks (such as the [1.1-Personalize](https://github.com/aws-samples/retail-demo-store/blob/master/workshop/1-Personalization/1.1-Personalize.ipynb) workshop). All other services should be regional or already use dynamic names.

The second phase will tackle using dynamic names throughout the project so that multiple deployments in the same account and same region will be supported.

alexchirayath:
Hi
@patrick-239  and @james-jory 
Multi-region deployment is still not available since IAM roles are still being used. 
I just tried it and reveived the following error:

`RetailDemoStore-orders-TaskRole already exists`

patrick-239:
How did you launch a stack (via button on GitHub page)?

alexchirayath:
@patrick-239 
I am trying to deploy it in the "Deploy for Development" mode but I am staging the resources and template in my own s3 buckets using the stage.sh script. However, I believe this would be an issue in all modes of deployment.
The issue is due to hard coding in resource names like roles as seen [here](https://github.com/aws-samples/retail-demo-store/blob/b7af99e0a46d68775228e8efd4870b759ccc6f4e/aws/cloudformation-templates/services/service/service.yaml#L120) and since IAM roles are account-level resources , any userdeploying in multiple regions could face this error

james-jory:
[...]

PR #70, opened 2020-08-28
Multiple deployment in one account

patrick-239:
*Issue #52  available:*

*Description of changes:*
All recourse use UID and now it is possible to deploy several demos in one account


patrick-239:
Will try to find a workaround for mark name

README.md
```diff
@@ -32,6 +32,8 @@ To get the Retail Demo Store running in your own AWS account, follow these instr
 
 ### AWS Deployment Options
 
+***IMPORTANT NOTE:** Please use only letters in Cloud Formation Stack name.*
```
alexchirayath, line 35: Just curious, is there a specific reason for using just letters?

aws/cloudformation-templates/_template.yaml
```diff
@@ -0,0 +1,302 @@
...
+    Type: String
+    Description: Amplitude API key for integrated product analytics and A/B testing results (optional).
+    NoEcho: true
+
+  OptimizelySdkKey:
+    Type: String
+    Description: Optimizely SDK key for experimentation (optional).
+    NoEcho: true
+
+  ParentStackName:
```
alexchirayath, line 93: Is there a need to pass this as an additional parameter?
Could we maybe use psuedo parameters like AWS::?
More info [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html)

README.md
patrick-239: unfortunately most of our services have a different naming rules. Problem with numbers comes from Lex (the support only letters) and as we define an IAM roles in CFN with resource based scope we have to define Lex name during deploy stage.

aws/cloudformation-templates/_template.yaml
patrick-239: Region could be received in any nested stack via pseudo parameter, but Parent stack name will be change every time when we use nested stack (we have many nested in nested ).

README.md
alexchirayath: Got it, thanks!

aws/cloudformation-templates/_template.yaml
alexchirayath: Makes sense.

alexchirayath (approved):

patrick-239:
Will cancel this pull request because:
1. Conflicts in different files. 
2. Macro name couldn't be unique

PR #81, opened 2020-09-12
Added demo guide

ShivaniMehendarge:
1. Added the modal for demo guide as per the prototype
2. Implemented all levels and content of the demo guide

Note: A few of the subsections do not have finalized content yet. It will be updated later.

*Issue #, if available:*

*Description of changes:*


src/web-ui/src/public/DemoGuideServiceDetails.vue
```diff
@@ -0,0 +1,120 @@
+<template>
+  <div class="demo-guide col-sm-12 col-md-9 col-lg-9 mb-4 about-service" style="padding:0">
+        <div v-for="subsection in subsections" v-bind:key=subsection.name style="padding:1em;" :style="{display: displaysection === subsection.name ? 'block' : 'none' }">
```
alexchirayath, line 3: Any specifc reason to use styles inline here?
Can we move all the styling to the bottom of the file under <style scoped>?

src/web-ui/src/public/components/DemoGuide.vue
```diff
@@ -0,0 +1,220 @@
...
+      this.showUseCases = false;
+      this.showPersonalizeTopics = false;
+    },
+    showDemoGuideLevel3: function() {
+      this.showguide = !this.showguide;
+      this.showguidedetails = !this.showguidedetails;
+    },
+    hideDemoGuideLevel3: function() {
+      this.showguide = !this.showguide;
+      this.showguidedetails = !this.showguidedetails;
```
alexchirayath, line 118: I see some members following camel case (eg: showUseCases) and some members that don't (eg: showguidedetails)
Is there a reason not to follow the same pattern?

ShivaniMehendarge: No reason. I can fix it.

src/web-ui/src/public/DemoGuideServiceDetails.vue
ShivaniMehendarge: It would actually create a lot more classes/divisions since many sections require a specific style - hence preferred putting it inline instead of creating that many classes.

james-jory (changes requested):
There is a lot of inline styles in the templates. These should be replaced with existing Bootstrap classes or scoped template styles.

src/web-ui/src/public/DemoGuideServiceDetails.vue
james-jory: How many are we talking about? I see many cases where inline styles can be replaced with Bootstrap classes. Also the inconsistency of using scoped template CSS with inline styling defeats the purpose of CSS and makes the templates harder to read and maintain.

ShivaniMehendarge: Makes sense. I will try to reduce as many as possible.

alexchirayath (approved):
LGTM

src/web-ui/src/public/components/DemoGuide.vue
```diff
@@ -0,0 +1,220 @@
...
+}
+.color-amazon-orange {
+  color:#e88b01
+}
+.no-padding {
+   padding: 0;
+}
+.no-margin {
+  margin: 0 ;
+}
```
james-jory, line 219: Consider using Bootstrap `p-0` and `m-0` classes instead. Same comment applies to all other uses of `no-padding` and `no-margin` throughout PR.

james-jory (changes requested):
Thank you for making the changes to the PR.

It appears that much of the CSS is duplicated across the `DemoGuide*.vue` components. For example, the `modal` class looks to be the same across all three. I think these should be factored out to either a global CSS file or into mixins. This will improve reuse and maintainability.

ShivaniMehendarge:
Closing this PR since @kazbaig has opened a new one (#124 )

PR #82, opened 2020-09-15
Add optimizely container params

manbearshark:
*Issue #, if available:*

Optimizely SDK key parameters were not being set for the Recommendations service

*Description of changes:*

Added SSM parameter and permissions to the service role.

### DIFF
--- a/aws/cloudformation-templates/services/service/service.yaml
+++ b/aws/cloudformation-templates/services/service/service.yaml
@@ -113,7 +113,15 @@ Resources:
               Service: ecs-tasks.amazonaws.com
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy
-
+      Policies:
+        - PolicyName: others
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              - Effect: Allow
+                Action:
+                  - ssm:GetParameters
+                Resource: !Sub 'arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/retaildemostore-*'
   TaskRole:
     Type: AWS::IAM::Role
     Properties:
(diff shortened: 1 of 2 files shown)


aws/cloudformation-templates/services/service/service.yaml
```diff
@@ -113,7 +113,15 @@ Resources:
               Service: ecs-tasks.amazonaws.com
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy
-
+      Policies:
+        - PolicyName: ReadRetailDemoStoreSSMParams
```
james-jory, line 117: Since this is only needed for accessing secure strings, consider renaming PolicyName so it's more clear. Otherwise it's potentially confusing why the TaskRole doesn't already cover this. Alternatively, add a CFN comment to make it clear why this is needed on the execution role. Thinking about maintenance and potential for something thinking it's not needed and removing it in the future.

aws/cloudformation-templates/services/service/service.yaml
```diff
@@ -207,6 +215,9 @@ Resources:
           Image: !Ref ContainerImage
           Essential: true
           Memory: 256
+          Secrets:
+            - Name: OPTIMIZELY_SDK_KEY
+              ValueFrom: !Sub 'arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/retaildemostore-optimizely-sdk-key'
```
james-jory, line 220: Does this work with this SSM parameter being created and value set via CloudFormation template parameter? Didn't think secure strings could be created and initialized via CFN. 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-parameter.html

james-jory (changes requested):

james-jory (approved):

Issue #80, opened 2020-09-08
Incorrect copy destination in CopyImagesLambdaFunction


alexchirayath:
I recently set up a deployment of the retail demo store using my own staging s3 buckets via the stage.sh script.
While the solution was deployed successfully, I am unable to see the images on the website.
On further researching, it seems that there is a bug in the CopyImagesLambdaFunction code.
The lambda function copies the files under **${RootURL}/${ResourceBucketRelativePath}images/** whereas the website is looking for the data under **${RootURL}/images/**

Proposed solution:  
The target object keys for the images needs to be updated to store in the root directory itself.
Change the logic in the lambda function would probably be [here](The target object keys for the images needs to be updated to store in the root directory itself.)

BastLeblanc:
fix in PR #83 please review :)

BastLeblanc:
updated fix in PR #84 please review :)

BastLeblanc:
closing this issue as #84 was merged

PR #83, opened 2020-09-15
Copy images lambda function fix

BastLeblanc:
*Issue #, if available:* #80 

*Description of changes:*
changed CopyImagesLambdaFunction to copy the images from sourcebucket/resourcebucket/images to targetbucket/images to match the rest of the configurations

### DIFF
--- a/aws/cloudformation-templates/web-ui-pipeline.yaml
+++ b/aws/cloudformation-templates/web-ui-pipeline.yaml
@@ -139,7 +139,12 @@ Resources:
 
                 for obj in source_bucket.objects.filter(Prefix=source_path):
                     source = { 'Bucket': source_bucket_name, 'Key': obj.key }
-                    target = target_bucket.Object(obj.key)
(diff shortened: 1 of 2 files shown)


aws/cloudformation-templates/web-ui-pipeline.yaml
```diff
@@ -139,7 +139,12 @@ Resources:
 
                 for obj in source_bucket.objects.filter(Prefix=source_path):
                     source = { 'Bucket': source_bucket_name, 'Key': obj.key }
-                    target = target_bucket.Object(obj.key)
+                    if source_path != "images":
```
alexchirayath, line 142: Would it be a better check to remove the relative path exists in the source path rather than checking if it the string "images"?
This would then work even if the "images" folder was renamed/restructured

BastLeblanc: that's a very good suggestion

PR #84, opened 2020-09-16
resource_bucket_path in CopyImagesLambdaFunction

BastLeblanc:
Issue #, if available: #80 

Description of changes:
changed CopyImagesLambdaFunction to copy the images from sourcebucket/resourcebucket/images to targetbucket/images to match the rest of the configurations

### DIFF
--- a/aws/cloudformation-templates/web-ui-pipeline.yaml
+++ b/aws/cloudformation-templates/web-ui-pipeline.yaml
@@ -130,6 +130,7 @@ Resources:
               source_bucket_name = event['ResourceProperties']['SourceBucket']
               source_path = event['ResourceProperties']['SourceBucketPath']
               target_bucket_name = event['ResourceProperties']['TargetBucket']
+              resource_bucket_path = event['ResourceProperties']['ResourceBucketRelativePath']
 
               if event['RequestType'] == 'Create' or event['RequestType'] == 'Update':
                 # Copy from source to target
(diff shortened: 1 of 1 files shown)


BastLeblanc:
integrated suggestion, checking on resourcepath directly, this will be more compatible (like works even if the resource path contains / itself)
Simply remove the resourcepath from the destination path to get images in TARGETBUCKET/images/...

aws/cloudformation-templates/web-ui-pipeline.yaml
```diff
@@ -139,7 +140,11 @@ Resources:
 
                 for obj in source_bucket.objects.filter(Prefix=source_path):
                     source = { 'Bucket': source_bucket_name, 'Key': obj.key }
-                    target = target_bucket.Object(obj.key)
+                    if len(resource_bucket_path)>0:
+                      # need to remove the resource_relative_path for the target images directory
+                      target = target_bucket.Object(obj.key[len(resource_bucket_path):])
```
alexchirayath, line 145: NIT: This works, I would prefer using resource_bucket_path.replace(resource_bucket_path, '')  rather than adding multiple conditions or using the length.

alexchirayath (approved):
LGTM
Just a nitpick comment

aws/cloudformation-templates/web-ui-pipeline.yaml
BastLeblanc: I wanted to avoid a "replace" because if resource_bucket_path contains somethings that is included later in the path it will break.
For example if customer used images/ as a resource path.. 


I could get this in one-liner, but I kept a if/else for clarity

PR #92, opened 2020-10-16
Remove dotenv

kazbaig:
*Issue #, if available:*

*Description of changes:*
Removes `dotenv` dependency as `vue-cli-service` already reads from the environment at build time. This branch is based off #91.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #93, opened 2020-10-16
Remove unnecessary deps

kazbaig:
*Issue #, if available:*

*Description of changes:*
Based off #92. Removes unnecessary dependencies. Had to update `aws-amplify` as it was causing an error during build that was being resolved by installing `lodash` explicitly on our end.


kazbaig:
Removing for now. Will revisit once the codebase reaches a more stable state.

PR #97, opened 2020-10-22
BugFix: Update csv path

alexchirayath:
When using RelativeResourceBucketPath, the CSV files are not written into the root of the bucket but under the ResourceBucketRelativePath. Fixing path with this commit

*Issue #, if available:*

*Description of changes:*
When using RelativeResourceBucketPath, the CSV files are not written into the root of the bucket but under the ResourceBucketRelativePath. Users deploying with a RelativeResourceBucketPath would face error in the Lambda function to precreate personalize campaign without this change.
Fixing the file  path with this commit


james-jory (approved):
Looks good. Thanks!

PR #98, opened 2020-10-23
Welcome page

kazbaig:
*Issue #, if available:*

*Description of changes:*
Adds welcome page to which the user is redirected if the `welcomePageVisited` flag in localStorage via `vuex` is false or missing and the user is not authenticated.


src/web-ui/src/public/Welcome.vue
```diff
@@ -0,0 +1,132 @@
...
+          <ul
+            class="service-list mb-0 pl-0 flex-grow-1 d-flex flex-column flex-lg-row justify-content-lg-around font-weight-bold"
+          >
+            <li v-for="service in services" :key="service.text" class="mb-2 mb-lg-0">
+              <ServiceBadge :text="service.text" :src="service.src"></ServiceBadge>
+            </li>
+          </ul>
+        </div>
+
+        <!-- make both links to go /auth until we have dedicated sign in and create account pages -->
```
alexchirayath, line 30: The Retail Demo Store has certain functionality that does not need sign in such viewing products/similar items , checkiing out as guest or accessing the demo guide. Do we need to add a "Skip this step" button?

kazbaig: Definitely makes sense to have that. We've opted to add a "skip this step" button to the roadmap and will make sure to add it before merging into the main branch.

alexchirayath (approved):
LGTM. Please wait until Jame's review to merge in

src/web-ui/src/public/Welcome.vue
```diff
@@ -0,0 +1,127 @@
...
+            *Have an account?<router-link to="/auth" class="sign-in ml-3 btn btn-link">Sign in</router-link>
+          </div>
+          <router-link to="/auth" class="create-account mt-3 mt-sm-0 ml-sm-3 btn btn-primary">Create an account</router-link>
+        </div>
+      </header>
+
+      <section class="info py-3 px-3 py-md-4 px-md-5">
+        <p>
+          *To keep the click-through data fresh, this demo resets every day at 11 pm EST. Accounts created more than 24
+          hours prior would be deleted as well. If your account has been deleted, go ahead and create a new one.
```
james-jory, line 42: I think we'll want to have this be more dynamic. Most deployments of this project will not be using the reset functionality at all. For example, if I deploy directly from GH. Also, and this is an open question for @alexchirayath, will the reset functionality be unique to the always-on mlsld deployment or something that can be optionally setup for any deployment? Finally, 11pm ET seems kind of early. Should that be configurable and then be dynamically displayed here?

src/web-ui/src/public/Welcome.vue
```diff
@@ -0,0 +1,127 @@
...
+        },
+        {
+          text: 'Amazon Pinpoint',
+          src: '/pinpoint.svg',
+        },
+        {
+          text: 'Amazon Lex',
+          src: '/lex.svg',
+        },
+      ],
```
james-jory, line 87: There are many more services than these used in the project. I know that the intent is to highlight the services that are related to customer experience but maybe we need to set that up better. I'm sure we'll be making some wording/marketing copy sweeps across all of these changes before it's all done so leaving this as a thought for now.

src/web-ui/src/config/localStorage/index.js
```diff
@@ -0,0 +1,22 @@
...
+    // handling the catch case below
+    // eslint-disable-next-line no-empty
+  } catch {}
+
+  if (defaultValue === undefined) return null;
+
+  localStorage.setItem(key, defaultValue);
+
+  return defaultValue;
+};
```
james-jory, line 22: Why not use the existing persisted state mechanism already in place in store.js?

james-jory (changes requested):

src/web-ui/src/public/Welcome.vue
alexchirayath: Thats a fair point. This probably should not be incorporated into the public demo.
The periodic reset mechanism is built into the deployment architecture (with CodeBuild and Step functions) which is outside of the retail demo store scope.

src/web-ui/src/public/Welcome.vue
kazbaig: That's fair. We'll be sure to revisit this issue before our deliverable date.

src/web-ui/src/config/localStorage/index.js
kazbaig: Moved this logic to use `vuex`.

src/web-ui/src/public/Welcome.vue
kazbaig: Removed the section for now.

james-jory (approved):

PR #100, opened 2020-10-24
Recommendations service performance improvement

james-jory:
*Description of changes:*

There has been a recent significant jump in latency to some of the resources in the Recommendations service. This seems to be tied to the Optimizely integration when Optimizely is not configured. Since all of the recommendations calls check for an active experiment for the feature and Optimizely is part of the logic path, something in the Optimizely library must be doing some slow-running checking when it's in this non-configured state. Added logic to only check for an Optimizely experiment when Optimizely is configured in the Retail Demo Store deployment.

### DIFF
--- a/src/recommendations/src/recommendations-service/experimentation/experiment_manager.py
+++ b/src/recommendations/src/recommendations-service/experimentation/experiment_manager.py
@@ -35,14 +35,14 @@ class ExperimentManager:
         """ Returns True if this environment is setup for running experiments """
         return self.__get_table()
 
-    def is_optimizey_configured(self):
+    def is_optimizely_configured(self):
         return optimizely_configured
 
     def get_active(self, feature):
         """ Returns the active experiment for the given feature """
         experiment = None
 
-        if self.is_optimizey_configured():
+        if self.is_optimizely_configured():
             config = optimizely_sdk.get_optimizely_config()
             if config:
                 if feature in config.features_map:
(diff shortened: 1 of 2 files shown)


src/recommendations/src/recommendations-service/experimentation/experiment_manager.py
```diff
@@ -35,25 +35,29 @@ def is_configured(self):
...
-                    experiment_key = list(experiment_keys)[0]
-                    experiment = optimizely_feature.experiments_map[experiment_key]
-                    data = {'id': experiment.id,
-                            'feature': feature,
-                            'name': experiment_key,
-                            'status': 'ACTIVE',
-                            'type': 'optimizely',
-                            'variations': []}
-                    return OptimizelyFeatureTest(None, **data)
+        if self.is_optimizey_configured():
```
yeeland, line 45:
```suggestion
        if self.is_optimizely_configured():
```

src/recommendations/src/recommendations-service/experimentation/experiment_manager.py
```diff
@@ -35,25 +35,29 @@ def is_configured(self):
         """ Returns True if this environment is setup for running experiments """
         return self.__get_table()
 
+    def is_optimizey_configured(self):
```
yeeland, line 38:
```suggestion
    def is_optimizely_configured(self):
```

yeeland (changes requested):

yeeland (approved):

manbearshark (approved):
LGTM

PR #111, opened 2020-11-07
Feature/adding abandoned cart ui trigger

alexchirayath:
*Issue #, if available:*

*Description of changes:*
Added functionality to trigger abandoned cart email via the button on the Cart page UI. 
Next: This would be merged in with #108 so that the button is only visible when signed in


james-jory (approved):
Changes look good.

PR #110, opened 2020-11-05
Product carousel

kazbaig:
*Issue #, if available:*

*Description of changes:*
Based off #108. Introduces a carousel UI for showing recommended products on `/product/:id` and `/`. Also changes home page to show both featured products and personalized recommendations on `/`.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #113, opened 2020-11-10
New Nav and footer

kazbaig:
*Issue #, if available:*

*Description of changes:*
Adds new nav and footer. Based off #110. Moved categories and cart data to `vuex` to avoid fetching the same data within multiple components on the same page.


james-jory (approved):
This is a massive PR with a lot of sweeping changes. It looks good and several improvements over the current implementation. Nice work! Will require a lot of integration testing and the merge into master is going to be a challenge.

kazbaig:
> This is a massive PR with a lot of sweeping changes. It looks good and several improvements over the current implementation. Nice work! Will require a lot of integration testing and the merge into master is going to be a challenge.

Thanks! It should look smaller after the prior PRs are merged. We can set up a time to discuss the PR at that point if you would like.

My goal is to start introducing tests as soon as we are code complete for the enhancement, though we would like to have been merging into master before then.

PR #116, opened 2020-11-16
Cold user improvements

james-jory:
*Description of changes:*

Modify cold user behavior as follows:

- When a new/cold user starts using the site, assign them a provisional user ID (UUID) that is tracked in local storage.
- As the new/cold user interacts with the site, include the provisional user ID in PutEvents calls to Personalize. Before we were just sending the session ID for cold users and not sending a user ID until the user signed in or created an account. This will allow cold user recommendations to start right away without having an account.
- Also in local storage, keep track of how many interactions have been sent to Personalize and automatically shift from default recommendation behavior to using recommendations from Personalize. For example, the home page will change from displaying "Featured" products (no user or provisional ID), to "Trending products" (provisional/new user but < 3 interactions), to "Inspired by your shopping trends" (after more than 2 events). The thinking is that popular products will be recommended to cold users with zero, 1, or 2 events but more than that and the recommendations should be more personalized.
- When a new/cold user creates an account, the provisional user ID will be sent to the Users service so that the same ID will be used as the user's actual user ID. This will keep continuity between the interactions before and after the user creates an account.


manbearshark (approved):
Looks pretty good to me.  This change should probably be made with the caveat that this type of identity scheme will result in multiple distinct user profiles being created across devices for users, and may result in multiple user identities for the same user if the browser decides to clear the local storage cache. In real life, this is a common problem as users tend to browse sites across mobile and laptop devices, which requires a solution to merge multiple temporary identifiers into a single user identifier.  The opposite approach is taken here, which is probably OK to showcase the semi-anonymous browsing use case but would have some serious potential issues if used in a production environment.  It would be cool if Personalize had a way to merge user-item data for the recommendations model once a user creates a permanent ID by signing up or by purchasing something.  As an alternative, the RDS users service could store a map of device-specific IDs mapped to a master user ID.  Lookups could  be allowed based on device ID, but return recommendations based on the user's ID if one is matched.

james-jory:
> Looks pretty good to me. This change should probably be made with the caveat that this type of identity scheme will result in multiple distinct user profiles being created across devices for users, and may result in multiple user identities for the same user if the browser decides to clear the local storage cache. 

I knew I'd get some great feedback from you! Really great points. The way I see it, this change doesn't really change things much from the prior implementation if anonymous users are periodically clearing local storage or using different devices without ever creating account. In fact, this solution is better in that these will users will at least get personalized recommendations. As for Personalize, these interactions will be ingested as separate sessions and contribute to SIMS and popularity but since any orphaned sessions will only impact Personalize and not the user tracking system.

> In real life, this is a common problem as users tend to browse sites across mobile and laptop devices, which requires a solution to merge multiple temporary identifiers into a single user identifier. The opposite approach is taken here, which is probably OK to showcase the semi-anonymous browsing use case but would have some serious potential issues if used in a production environment. 

I'd like to learn more about these issues! With this PR, if the same logical user browses the site anonymously using their phone and laptop, separate provisional user IDs will be created. If that user creates an account using one device and then signs into that account from the other device, both devices will start using the same user ID. So you're really only orphaning the interaction history from one of the devices from before the account was created. What am I missing?

> It would be cool if Personalize had a way to merge user-item data for the recommendations model once a user creates a permanent ID by signing up or by purchasing something. 

Personalize does this already today and in fact is what the RDS was using before this PR. That is, interactions for anonymous sessions were streamed via PutEvents with a sessionId only and no userId. Then, when that anonymous user creates an account, the next PutEvents calls have the same sessionId **and** now the userId. Personalize will merge those prior anonymous interactions for the same sessionId under the same user. This PR just accelerates this with a provisional user ID so we get cold recs immediately.

> As an alternative, the RDS users service could store a map of device-specific IDs mapped to a master user ID. Lookups could be allowed based on device ID, but return recommendations based on the user's ID if one is matched.

Or I could just use a CDP for this, right? ;-)

PR #131, opened 2020-12-03
Pr/requested fixes

AkramDweikat:
- Added flag to prevent CloudFormaiton deployment of IVS resources when default streams flag is set to Yes
- Move interactions generation into the right place

This work was done by Dae.mn Team.
  Co-authored-by: Akram Dweikat akram@akramdweikat.com
  Co-authored-by: Akram Dweikat akram.dweikat@dae.mn
  Co-authored-by: Damien Jade Duff damien.duff@dae.mn
  Co-authored-by: Joe Major joe.major@dae.mn

PR #132, opened 2020-12-04
turn off IVS for EE

patrick-239:
*Issue #, if available:*

*Description of changes:*
Added new parameter UseIVSStreams to prevent IVS lambdas deployment in EE. Previously used parameter UseDefaultIVSStreams was designed for another function.  

### DIFF
--- a/aws/cloudformation-templates/base/notebook.yaml
+++ b/aws/cloudformation-templates/base/notebook.yaml
@@ -37,7 +37,7 @@ Resources:
     Type: AWS::SageMaker::NotebookInstance
     Properties:
       NotebookInstanceName: !Sub ${Uid}
-      InstanceType: "ml.t2.medium"
+      InstanceType: "ml.t3.medium"
       RoleArn: !GetAtt ExecutionRole.Arn
       SubnetId: !Ref Subnet1
       SecurityGroupIds:
--- a/aws/cloudformation-templates/deployment-support.yaml
+++ b/aws/cloudformation-templates/deployment-support.yaml
@@ -88,7 +88,7 @@ Parameters:
   CustomizeRecommendationsFunctionArn:
     Type: String
 
-  UseDefaultIVSStreams:
+  UseIVSStreams:
     Type: String
     Default: Yes
 
@@ -106,8 +106,8 @@ Conditions:
     - 'Yes'
 
   CreateIVSResources: !Equals
-    - !Ref UseDefaultIVSStreams
-    - 'No'
+    - !Ref UseIVSStreams
+    - 'Yes'
 
 Resources:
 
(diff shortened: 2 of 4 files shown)


james-jory:
How does this change the current interim default behavior?

patrick-239:
Currently to use RDS for AWS managed workshop you have to use set UseDefaultIVSStreams = Yes and it will block IVS lambdas deployment. In a same time if someone deploy it for a demo or to explore demo he will see UseDefaultIVSStreams parameter in console and will select "Yes", but a results will be opposite - IVS lambdas will not be deployed. UseDefaultIVSStreams was designed to change configuration of IVS: to use default stream or custom from a resource folder. 

This is a reason why I want to introduce special parameter to control Lambda deployments without any influence on a IVS configuration.

patrick-239:
added one more commit: switching from t2 to t3 instance family for a SageMaker Jupiter instance.

damiendaemon:
There may be some confusion with this commit: https://github.com/aws-samples/retail-demo-store/commit/b4da87402487025b35dfcd00bbac6caf9f31525e - I might humbly suggest that `UseDefaultIVSStreams` should be "Yes" in regions that do not support IVS and so should default to "Yes" - so that the default streams are used and the demo does not attempt to deploy its own. I'd suggest reverting that. The following merged PR may also address some of the concerns raised in this PR: https://github.com/aws-samples/retail-demo-store/pull/131

damiendaemon:
Note first that as you surmised the intent of the `UseDefaultIVSStreams` is to allow the backend to pick up IVS streams from another, default, deployment of the streams (which are deployed permanently in the aws-retail account). Therefore when this flag is "**Yes**" there is no need to deploy any IVS resources in the current deployment. Therefore the intent of the following (merged into master) PR is to address this: https://github.com/aws-samples/retail-demo-store/pull/131 Here, if UseDefaultIVSStreams is set to "**Yes**" (that is, we request to use the default streams and not our own ones) the deployment is not going to deploy any IVS resources at all. We believe this is the intent of the current PR, but this is currently handled by the same flag.

Therefore, we also humbly suggest reverting the following commit and setting `UseDefaultIVSStreams` to "**Yes**" by default: https://github.com/aws-samples/retail-demo-store/commit/b4da87402487025b35dfcd00bbac6caf9f31525e - thereby, by default, no IVS resources will be deployed.  The previous issue (solved in the above-mentioned PR#131) resulted from the fact that the IVS resources were still being created, even when using the default streams.

Note further that the _Live_ tab in the UI will continue to work as before even if no IVS resources are deployed in the current deployment, as the streams are taken from the separate above-mentioned aws-retail-hosted permanent deployment. If it is wanted to remove the _Live_ tab altogether we would suggest adding a flag to top level CloudFormation, `web-ui-pipeline.yaml `and `gen_env.sh` to create a switch in the UI. However  we want to emphasise is unnecessary as the tab will work without problem without hosting your own streams.

We hope this has helped resolve this issue but of course do feel free to remind of any issues we may have missed.

patrick-239:
@damiendaemon thank you so mush for an explanation! Before that I had a wrong understanding what means "default stream". Now I totally agree, that UseDefaultIVSStreams is resolving an issue and my PR in not necessary.

PR #135, opened 2020-12-07
Updates to checkout page

jbt:
* Removed all address / billing inputs
 * Reorganised order summary box
 * Added trigger abandoned cart email button

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/web-ui/src/public/Checkout.vue
+++ b/src/web-ui/src/public/Checkout.vue
@@ -32,22 +32,6 @@
                 <ul class="list-group list-group-flush mb-3">
                   <li class="list-group-item p-1 d-flex justify-content-between">
                     <span>{{ this.cartQuantity }} item{{ this.cartQuantity === 1 ? '' : 's' }} in cart</span>
-                    <strong>${{ this.cartSubTotal.toFixed(2) }}</strong>
-                  </li>
-                  <li class="list-group-item p-1 d-flex justify-content-between border-top-0">
-                    <span>Shipping & handling:</span>
-                    <strong>${{ this.cartShippingRate.toFixed(2) }}</strong>
-                  </li>
-                  <li class="list-group-item p-1 d-flex justify-content-between">
-                    <span>Total before tax:</span>
-                    <strong>${{ this.cartTotalBeforeTax.toFixed(2) }}</strong>
-                  </li>
-                  <li class="list-group-item p-1 d-flex justify-content-between border-top-0">
-                    <span>Tax to be collected:</span>
-                    <strong>${{ this.cartTaxRate.toFixed(2) }}</strong>
-                  </li>
-                  <li class="list-group-item p-1 d-flex justify-content-between border-bottom-0">
-                    <span>Order total:</span>
                     <strong>${{ this.cartTotal.toFixed(2) }}</strong>
                   </li>
                 </ul>
@@ -55,7 +39,7 @@
               </div>
             </div>
 
-            <button v-if="pinpointEnabled && user" v-on:click="triggerAbandonedCartEmail" class="abandoned-cart-btn btn btn-primary btn-lg m-4">
+            <button v-if="true || pinpointEnabled && user" v-on:click="triggerAbandonedCartEmail" class="abandoned-cart-btn btn btn-primary btn-lg m-4">
               Trigger Abandoned Cart email
             </button>
           </div>
@@ -63,17 +47,18 @@
             <div class="alert text-center ml-0 not-real-warning" v-if="showCheckout == true">This storefront is not real.<br/>Your order will not be fulfilled.</div>
 
             <form>
-              <div class="row">
+              <div class="d-flex">
                 <h5 class="p-4 col-md-4 bg-light font-weight-bold">Shipping Address</h5>
                 <div class="col-md-8">
-                  <p class="mb-1">Joe Doe</p>
-                  <p class="mb-1">2730 Sample address Ave.</p>
-                  <p class="mb-1">Seattle, WA 98109</p>
+                  <p class="mb-1 font-weight-bold">{{order.shipping_address.first_name}} {{order.shipping_address.last_name}}</p>
+                  <p class="mb-1">{{order.shipping_address.address1}}</p>
+                  <p class="mb-1" v-if="order.shipping_address.address2">{{order.shipping_address.address2}}</p>
+                  <p class="mb-1">{{order.shipping_address.city}}, {{order.shipping_address.state}} {{order.shipping_address.zipcode}}</p>
                 </div>
               </div>
               <hr class="mb-4">
 
-              <div class="row">
+              <div class="d-flex">
                 <h5 class="p-4 col-md-4 bg-light font-weight-bold">Payment</h5>
                 <div class="col-md-8">
                   <p class="mb-1">VISA ending in 0965</p>
@@ -97,7 +82,7 @@
 </template>
 
 <script>
-import {mapState} from 'vuex';
+import {mapGetters, mapState} from 'vuex';
 
 import AmplifyStore from '@/store/store'
 
@@ -114,7 +99,7 @@ const OrdersRepository = RepositoryFactory.get('orders')
 export default {
   name: 'Checkout',
   components: {
-    Layout,
+    Layout
   },
   props: {
   },
@@ -162,6 +147,19 @@ export default {
         this.order = {}
         this.order.promo_code = ""
       }
+
+      if (!this.order.billing_address) {
+        this.order.billing_address = this.order.shipping_address = {
+          first_name: 'Joe',
+          last_name: 'Doe',
+          email: this.user.email,
+          address1: '2730 Sample address Ave.',
+          address2: '',
+          city: 'Seattle',
+          state: 'WA',
+          zipcode: '98109'
+        }
+      }
     },
     signIn () {
       this.$router.push('/auth')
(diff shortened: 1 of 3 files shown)


src/web-ui/src/public/Checkout.vue
```diff
@@ -15,151 +15,88 @@
...
+                  </li>
+                </ul>
+                <button class="btn btn-primary btn-lg btn-block" v-on:click="submitOrder">Place your Order</button>
               </div>
-            </form>
+            </div>
+
+            <button v-if="pinpointEnabled && user" v-on:click="triggerAbandonedCartEmail" class="abandoned-cart-btn btn btn-primary btn-lg m-4">
+              Trigger Abandoned Cart email
+            </button>
```
james-jory, line 65: Should this be on the cart view instead?

src/web-ui/src/public/Checkout.vue
```diff
@@ -15,151 +15,88 @@
...
-                  <label for="zip">Zip</label>
-                  <input type="text" class="form-control" id="zip" v-model="order.billing_address.zipcode" placeholder="" required>
-                  <div class="invalid-feedback">
-                    Zip code required.
-                  </div>
+                <h5 class="p-4 col-md-4 bg-light">Shipping Address</h5>
+                <div class="col-md-8">
+                  <p class="mb-1">Joe Doe</p>
+                  <p class="mb-1">2730 Sample address Ave.</p>
+                  <p class="mb-1">Seattle, WA 98109</p>
```
james-jory, line 76: The shipping details for the emulated shopper are still being sent to the orders service so we should display the details for the emulated shopper. They should be available in the `order.shipping_address` and `order.billing_address`.

james-jory (commented):
Looks good. A couple comments.

src/web-ui/src/public/Checkout.vue
alexchirayath: One additional comment is that, this page should have RO forms so as to avoid users accidentally providing real data for shipping/billing

src/web-ui/src/public/Checkout.vue
alexchirayath: +1

jbt: Ah for this one I wasn't sure - the designs I saw had this on _both_ cart and checkout views so I brought it in additionally here

kazbaig: According to the designs, it should be on both pages

src/web-ui/src/public/Checkout.vue
```diff
@@ -262,6 +187,15 @@ export default {
...
+      if (this.cart && this.cart.items.length > 0 ){
+        const cartItem = await this.getProductByID(this.cart.items[0].product_id)
+        AnalyticsHandler.recordAbanonedCartEvent(this.user,this.cart,cartItem)
+      }
+      else{
+        console.error("No items to export")
+      }
     }
   },
   computed: {
```
kazbaig, line 201: Let's use the existing getters in `vuex` to fetch cart quantities. You'll notice that some calculations like tax are not available - these have been removed on the cart page and can also be removed here.

src/web-ui/src/public/Checkout.vue
```diff
@@ -15,151 +15,88 @@
...
-                </div>
-                <div class="col-md-3 mb-3">
-                  <label for="zip">Zip</label>
-                  <input type="text" class="form-control" id="zip" v-model="order.billing_address.zipcode" placeholder="" required>
-                  <div class="invalid-feedback">
-                    Zip code required.
-                  </div>
+                <h5 class="p-4 col-md-4 bg-light">Shipping Address</h5>
+                <div class="col-md-8">
+                  <p class="mb-1">Joe Doe</p>
```
kazbaig, line 74: This should be bold

src/web-ui/src/public/Checkout.vue
```diff
@@ -15,151 +15,88 @@
...
+                  <li class="list-group-item p-1 d-flex justify-content-between border-top-0">
+                    <span>Tax to be collected:</span>
+                    <strong>${{ this.cartTaxRate.toFixed(2) }}</strong>
+                  </li>
+                  <li class="list-group-item p-1 d-flex justify-content-between border-bottom-0">
+                    <span>Order total:</span>
+                    <strong>${{ this.cartTotal.toFixed(2) }}</strong>
+                  </li>
+                </ul>
+                <button class="btn btn-primary btn-lg btn-block" v-on:click="submitOrder">Place your Order</button>
```
kazbaig, line 59: Let's use the same colors as the Checkout button on `/cart` for this one

kazbaig:
We're missing the "Back to shopping cart" link on the top left. You can use the `previousPageLinkProps` prop on `Layout` to achieve this.

kazbaig:
While we're making updates, let's also use the `LoadingFallback` component instead of hand-rolled markup for the loading indicator.

src/web-ui/src/public/Checkout.vue
```diff
@@ -15,151 +15,88 @@
...
-              <div class="mb-3">
-                <label for="address">Address</label>
-                <input type="text" class="form-control" id="address" v-model="order.billing_address.address1" placeholder="1234 Main St" required>
-              </div>
-
-              <div class="mb-3">
-                <label for="address2">Address 2 <span class="text-muted">(Optional)</span></label>
-                <input type="text" class="form-control" id="address2" v-model="order.billing_address.address2" placeholder="Apartment or suite">
-              </div>
+            <div class="alert alert-secondary text-center ml-0" v-if="showCheckout == true">This storefront is not real.<br/>Your order will not be fulfilled.</div>
```
kazbaig, line 68: This div should be light blue. Let's add `blue-100: #f1faff;` to the tokens css file and use that as the background

src/web-ui/src/public/Checkout.vue
```diff
@@ -15,157 +10,79 @@
...
+          <div class="col-md-4 order-md-2">
+            <div class="card p-1">
+              <div class="card-body">
+                <h4 class="d-flex justify-content-between align-items-center mb-3 card-title text-muted">
+                  Order Summary
+                </h4>
+                <ul class="list-group list-group-flush mb-3">
+                  <li class="list-group-item p-1 d-flex justify-content-between">
+                    <span>{{ this.cartQuantity }} item{{ this.cartQuantity === 1 ? '' : 's' }} in cart</span>
+                    <strong>${{ this.cartTotal.toFixed(2) }}</strong>
```
kazbaig, line 35: I don't think this needs to be a list. Let's also remove the top/bottom borders on this line item and increase the font so it's the only piece of data.

src/web-ui/src/public/Checkout.vue
```diff
@@ -15,157 +10,79 @@
...
+                  </li>
+                </ul>
+                <button class="checkout-btn btn btn-outline-dark btn-block btn-lg btn-block" v-on:click="submitOrder">Place your Order</button>
               </div>
-            </form>
+            </div>
+
+            <button v-if="true || pinpointEnabled && user" v-on:click="triggerAbandonedCartEmail" class="abandoned-cart-btn btn btn-primary btn-lg m-4">
```
kazbaig, line 42: Looks like the `true` is a a leftover placeholder

kazbaig, line 42: This button should match the width and positioning of the confirmation button above it.

src/web-ui/src/public/Checkout.vue
```diff
@@ -15,157 +10,79 @@
...
+                <h4 class="d-flex justify-content-between align-items-center mb-3 card-title text-muted">
+                  Order Summary
+                </h4>
+                <ul class="list-group list-group-flush mb-3">
+                  <li class="list-group-item p-1 d-flex justify-content-between">
+                    <span>{{ this.cartQuantity }} item{{ this.cartQuantity === 1 ? '' : 's' }} in cart</span>
+                    <strong>${{ this.cartTotal.toFixed(2) }}</strong>
+                  </li>
+                </ul>
+                <button class="checkout-btn btn btn-outline-dark btn-block btn-lg btn-block" v-on:click="submitOrder">Place your Order</button>
```
kazbaig, line 38: The "O" shouldn't be capitalized here

src/web-ui/src/public/Checkout.vue
```diff
@@ -15,157 +10,79 @@
...
-                </div>
+          <div class="col-md-4 order-md-2">
+            <div class="card p-1">
+              <div class="card-body">
+                <h4 class="d-flex justify-content-between align-items-center mb-3 card-title text-muted">
+                  Order Summary
+                </h4>
+                <ul class="list-group list-group-flush mb-3">
+                  <li class="list-group-item p-1 d-flex justify-content-between">
+                    <span>{{ this.cartQuantity }} item{{ this.cartQuantity === 1 ? '' : 's' }} in cart</span>
```
kazbaig, line 34: We don't use `this` within vue templates

kazbaig:
Since there isn't a cart subtotal anymore, I'm getting a console error in the analytics handler file. Can you please remove the subtotal being passed to `AnalyticsHandler.checkoutStarted` from both the handler itself and the function call?

src/web-ui/src/public/Checkout.vue
```diff
@@ -15,157 +10,79 @@
...
-                </div>
-                <div class="col-md-3 mb-3">
-                  <label for="zip">Zip</label>
-                  <input type="text" class="form-control" id="zip" v-model="order.billing_address.zipcode" placeholder="" required>
-                  <div class="invalid-feedback">
-                    Zip code required.
-                  </div>
+            <form>
+              <div class="d-flex">
+                <h5 class="p-4 col-md-4 bg-light font-weight-bold">Shipping Address</h5>
```
kazbaig, line 51: We can vertically center this and the Payment heading below with `d-flex align-items-center`

src/web-ui/src/public/Checkout.vue
```diff
@@ -15,157 +10,79 @@
...
+                    <strong>${{ this.cartTotal.toFixed(2) }}</strong>
+                  </li>
+                </ul>
+                <button class="checkout-btn btn btn-outline-dark btn-block btn-lg btn-block" v-on:click="submitOrder">Place your Order</button>
               </div>
-            </form>
+            </div>
+
+            <button v-if="true || pinpointEnabled && user" v-on:click="triggerAbandonedCartEmail" class="abandoned-cart-btn btn btn-primary btn-lg m-4">
+              Trigger Abandoned Cart email
```
kazbaig, line 43: Since this button is looking pretty funky on medium width screens before the mobile layout kicks in, let's either use auto-sizing similar to the cart page for this column or use the mobile layout sooner.

kazbaig:
Seeing `Error in created hook (Promise/async): "TypeError: Cannot read property 'email' of null` when loading the page unauthenticated

src/web-ui/src/public/Checkout.vue
kazbaig, line 35: We should use `formattedCartTotal` which we fetched from `vuex` but aren't using right now here instead of formatting by hand.

src/web-ui/src/public/Checkout.vue
jbt: Ah yep I thought I'd reset that back before pushing, good spot

src/web-ui/src/public/Checkout.vue
jbt: Yep absolutely now we've removed the extra items it doesn't make sense to be a list any more

src/web-ui/src/public/Checkout.vue
jbt: Ah yes this was copypasted from where it was used previously, will simplify

jbt:
Ok hopefully most of those should now be addressed - I'm not seeing the `first_name` error myself but I suspect that was only happening because of the previous `email` error which is now fixed

kazbaig:
Look good now :)

james-jory (approved):

PR #137, opened 2020-12-07
Feature/user filtering

alexchirayath:
*Issue #, if available:*

*Description of changes:*
Added in changes to support the new UX-

1) GetRandomUser
This returns a random shopper persona which can be used by the logged in user

2) GetFilteredUser
Returns a shopper persona matching the inputed primary interest(persona) and agerange (Age range is a concept that was defined on the UI and the corresponding backend mapping function has been created)

3) ClaimUser
This marks all shopper personas that have been 'claimed' or are 'used' by other logged in users. These marked users can then be excluded from the Filter shopper get random user APIs so that no two logged in users can have the same shopper persona thereby reducing the chance of providing polluted results. As dicussed with @james-jory  and others, this is a best efforts temporary update. True idempotent and transnational based user claiming and filtering would be provided by a database service like DynamoDB

Next steps: Kaz to send out an PR with the frontend changes to support the new Shopper select and filtering


src/users/src/users-service/handlers.go
```diff
@@ -122,7 +122,47 @@ func UserShowByIdentityId(w http.ResponseWriter, r *http.Request) {
...
+func ClaimUser(w http.ResponseWriter, r *http.Request) {
+	
+	enableCors(&w)
+	if (*r).Method == "OPTIONS" {
+		w.WriteHeader(http.StatusOK)
+		return
+	}
+	r.ParseForm()
+	identityId := r.Form.Get("identityId")
+	userId, err := strconv.Atoi(r.Form.Get("userId"))
```
james-jory, line 135: Sending in payloads using `application/x-www-form-urlencoded` content type breaks precedence across the other APIs where we're using JSON payloads. I think we should stay consistent or ideally, in this specific case, be more RESTful and use `POST /users/id/{userID}/claim`. I realize this starts to get into subjective territory on RESTful API design but it feels like putting the `claim` verb on the end of the exiting `/users/id/{userID}` is more intuitive.

src/users/src/users-service/routes.go
```diff
@@ -47,6 +49,30 @@ var routes = Routes{
...
+	Route{
+		"GetFilteredUser",
+		"GET",
+		"/users/filter/",
+		GetFilteredUser,
+	},
+	Route{
+		"GetRandomUser",
+		"GET",
+		"/users/getRandomUser/",
```
james-jory, line 61: I'd prefer something like `GET /users/random?count=X` for consistency where count defaults to `1`.

src/users/src/users-service/routes.go
```diff
@@ -47,6 +49,30 @@ var routes = Routes{
 		"/users/identityid/{identityID}",
 		UserShowByIdentityId,
 	},
+	Route{
+		"GetFilteredUser",
+		"GET",
+		"/users/filter/",
```
james-jory, line 55: Can we change this to `GET /users` or `GET /users/all` with optional query string parameters that can be used to filter based on primary persona, age range, etc?

src/users/src/users-service/routes.go
```diff
@@ -47,6 +49,30 @@ var routes = Routes{
...
+	Route{
+		"GetRandomUser",
+		"GET",
+		"/users/getRandomUser/",
+		GetRandomUser,
+	},
+	Route{
+		"ClaimUser",
+		"POST",
+		"/users/claim/",
```
james-jory, line 67: I'd prefer `PUT /users/{userID}/claim` to keep it tied to a specific user resource. The identityId can be passed as a JSON payload.

james-jory (changes requested):

src/users/src/users-service/routes.go
alexchirayath: Makes sense. Updated

src/users/src/users-service/handlers.go
alexchirayath: Makes sense. Also, I just realzied that we dont actually need the identityID to be passed since we just want to keep track on whether the user is claimed or not. Updated to use PUT /users/id/{userID}/claim

src/users/src/users-service/routes.go
alexchirayath: I updated the endpoint to users/unclaimed with params of ageRange,primaryPersona and count.
I'd like to keep this as is and not merge it into users/ users/all since this API is built to only return users for the shopper-select criteria for a very specific use.
The rest of the comments on the PR, I agree with and have made updates.

src/users/src/users-service/routes.go
alexchirayath: Actually, we dont really need the identityId to be passed at all since we just want to claim a shopper and there is no need to support the entire audit of which identity is used by which shopper. Updated the method

james-jory (approved):

PR #138, opened 2020-12-09
Remove username signup

ShivaniMehendarge:
*Issue #, if available:*

*Description of changes:*
Removed the username field as per the UX/UI mocks. For the cognito user, display email id instead of username on UI, and use email to signup/ signin.

### DIFF
--- a/src/web-ui/src/router/index.js
+++ b/src/web-ui/src/router/index.js
@@ -72,7 +72,7 @@ AmplifyEventBus.$on('authState', async (state) => {
       storeUser = data
     }
     else {
-      const { data } = await UsersRepository.getUserByUsername(cognitoUser.username)
+      const { data } = await UsersRepository.getUserByUsername(cognitoUser.attributes.email)
       storeUser = data
     }
 
@@ -83,7 +83,7 @@ AmplifyEventBus.$on('authState', async (state) => {
       console.log('store user does not exist for cognito user... creating on the fly')
       let identityId = credentials ? credentials.identityId : null;
       let provisionalUserId = AmplifyStore.getters.personalizeUserID;
-      const { data } = await UsersRepository.createUser(provisionalUserId, cognitoUser.username, cognitoUser.attributes.email, identityId)
+      const { data } = await UsersRepository.createUser(provisionalUserId, cognitoUser.attributes.email, cognitoUser.attributes.email, identityId)
       storeUser = data
     }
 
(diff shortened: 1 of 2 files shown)


src/web-ui/src/router/index.js
```diff
@@ -72,7 +72,7 @@ AmplifyEventBus.$on('authState', async (state) => {
       storeUser = data
     }
     else {
-      const { data } = await UsersRepository.getUserByUsername(cognitoUser.username)
+      const { data } = await UsersRepository.getUserByUsername(cognitoUser.attributes.email)
```
james-jory, line 75: Since username is going away, the Users service endpoints should change as well from `getUserByUsername` to `getUserByEmail`.

src/web-ui/src/router/index.js
```diff
@@ -83,7 +83,7 @@ AmplifyEventBus.$on('authState', async (state) => {
       console.log('store user does not exist for cognito user... creating on the fly')
       let identityId = credentials ? credentials.identityId : null;
       let provisionalUserId = AmplifyStore.getters.personalizeUserID;
-      const { data } = await UsersRepository.createUser(provisionalUserId, cognitoUser.username, cognitoUser.attributes.email, identityId)
+      const { data } = await UsersRepository.createUser(provisionalUserId, cognitoUser.attributes.email, cognitoUser.attributes.email, identityId)
```
james-jory, line 86: This change creates tech debt. Since the username field is going away, the `createUser` function for the repository and the API call to the create the user in the Users service should change to no longer take a username.

james-jory (changes requested):
These changes only partially address removing username and replacing username with email as a logical key for lookups.

The username must also be removed from the Users service, the `/users/username/{username}` endpoint needs to be changed in the Users service, and the users data generator needs to be changed to remove username from the dataset.

src/web-ui/src/router/index.js
ShivaniMehendarge: After discussing with the team about the tradeoffs of having the username field separately, we would like to avoid the risk of impacting the existing functionality and workshop or creating tech debt for the future.

ShivaniMehendarge:
After discussing with the team about the tradeoffs of having the username field separately, we would like to avoid the risk of impacting the existing functionality and workshop or creating tech debt for the future. Hence closing this PR.

PR #134, opened 2020-12-04
Demo walkthrough

kazbaig:
*Issue #, if available:*

*Description of changes:*
Based off #124.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #142, opened 2020-12-10
Shopper profile select

kazbaig:
Based off #134.

Changes:
- Add new method for switching shopper profiles. It is now accessible as a model opened using the "Switch shopper" option in the revamped `UserDropdown`. This allows the user to select a random shopper, switch back to the authenticated user's default profile, switch to a profile matching provided age range and primary interest.
- Detect if a sign in event is for a new account. If so, redirect to `shopper-select`, an in-page version of the shopper selection flow.
- In addition to showing when the home page is first visited, show the demo walkthrough on all visits of the home page after logging in with a new account.
- Remove `/profile` page as it is made redundant by the new shopper selection flow.
- Change `CategoryDropdown` to `ExploreDropdown` to convey that it contains more than just categories. Also adds back previously removed "Featured" pseudo-category. 
- Add logic for cleaning up background color injected by `Layout` as we now have more than one layout component.
- Convert `PoweredByAWS` SVG to a component in order to reuse when we need different text colors


src/web-ui/src/partials/AppModal/DemoWalkthrough/DemoWalkthroughPageLayout.vue
```diff
@@ -0,0 +1,240 @@
...
+import { mapActions, mapState } from 'vuex';
+
+import { APP_MODAL_ID } from '../config';
+
+export default {
+  name: 'DemoWalkthroughPageLayout',
+  props: {
+    showNav: { type: Boolean, default: true },
+    isLast: { type: Boolean, default: false },
+    // Side: 'TOP' | 'RIGHT' | 'BOTTOM' | 'LEFT'
```
alexchirayath, line 54: Lets remove the commented code if not needed

kazbaig: It's documentation for the types, not commented code

src/web-ui/src/partials/Navigation/UserDropdown/UserDropdown.vue
```diff
@@ -43,9 +63,38 @@ export default {
...
       },
+      gender() {
+        if (!this.user) return null;
+
+        switch (this.user.gender) {
+          case 'M':
+            return 'Male';
+          case 'F':
+            return 'Female';
+        }
```
james-jory, line 74: Need to account for edge cases. Gender (and persona and age for that matter) is not set on new users created through the sign in process where either a shopper is not emulated or the user chooses to fall back to their default profile. This code and the UI design should be defensive to those cases. Also, a user can technically be added via the Users service API without a gender or persona.

src/web-ui/src/partials/ShopperSelect/pages/SelectShopper.vue
```diff
@@ -0,0 +1,144 @@
...
+            <select class="form-control" id="primary-interest" v-model="primaryInterest">
+              <option value="">Select primary interest</option>
+              <option value="footwear">Footwear</option>
+              <option value="housewares">Housewares</option>
+              <option value="apparel">Apparel</option>
+              <option value="jewelry">Jewelry</option>
+              <option value="beauty">Beauty</option>
+              <option value="electronics">Electronics</option>
+              <option value="accessories">Accessories</option>
+              <option value="outdoors">Outdoors</option>
```
james-jory, line 38: These should be loaded from the `/categories/all` endpoint on the Products service in case categories change and when we add support for deploying with custom datalogs.

james-jory (changes requested):

src/web-ui/src/partials/Navigation/UserDropdown/UserDropdown.vue
kazbaig: This bit of code is only relevant to the scenario in which a shopper profile is selected (`v-if="user.persona"` on line 13). In that scenario, persona, age, and gender should be defined if I'm not mistaken. Do we still need to make the assumptions you're describing?

alexchirayath: I believe this is only for the shopper profiles and not for other users. Hence, probably this case wont come up

src/web-ui/src/partials/ShopperSelect/pages/SelectShopper.vue
alexchirayath: Agreed, Thats a good point.

kazbaig: Agreed that they should be fetched dynamically. They're hard coded right now because of the status of the dataset updates, which is that products with new categories have been introduced but shoppers with those categories as their persona have not.

Assuming we'll hold off on merging the dataset improvements until shopper profiles with the new categories as persona are added, I can roll this back to dynamically fetching categories.

kazbaig: Done

src/web-ui/src/partials/Navigation/UserDropdown/UserDropdown.vue
james-jory: Emulating one of the fictitious shoppers/users where persona, gender, and age are reliably defined is one demo path. Using your own profile (created when you sign up) is another demo path. In that case, since there are not fields on the sign up page for persona, gender, and age, those fields in the User object will be null. This is the cold user demo path and the UI needs to be tolerant of these fields being null. Correct me if I'm wrong but it's my understanding that this dropdown applies for both demo paths.

Having logic in the `gender` function assume/rely on a `v-if` in the template to not generate a runtime error seems unnecessarily brittle to me but I won't fight you on it. I just want to make sure that the emulated shopper *and* cold user demo paths will work as expected.

kazbaig: The reasoning for doing it this way was that the `v-if` directive functions similar to a typeguard in TS. Once we switch to TS, we will have clearer guarantees about the safety of trying to access properties. In the meantime, I've made the gender calculation safer just in case. If some edge case (which I don't think is possible anyway) takes place, the UI would fail gracefully now.

james-jory (approved):

PR #143, opened 2020-12-12
added maxResults=100 parameter to list-schemas code

manbearshark:
*Issue #, if available:*

The recovery code that checks for existing schemas did not take into account that there may be more than the default number of results from the list-schemas API call, this broke the Personalize notebook for me.

*Description of changes:*

Added a maxResults=100 parameter to the list-schemas call and a print statement to notify workshop participants that an existing schema is being used.


james-jory (approved):
We should change to paginators at some point but this should be fine for most cases.

PR #149, opened 2020-12-14
Update auth screen to be own component and render conditional message

jbt:
*Issue #, if available:*

*Description of changes:*

* Moves authenticator into own component now using SecndaryLayout
* Minor styling tweaks to remove box shadow and use consistent link color, but not introduce any further CSS hacking inside authenticator
* Watches the display state of the authenticator form and conditionally renders notice above the sign up form

### DIFF
--- a/src/web-ui/src/public/Auth.vue
+++ b/src/web-ui/src/public/Auth.vue
@@ -1,0 +1,148 @@
+<template>
+  <secondary-layout>
+    <div v-if="showingSignUp" class="sign-up-notice">
+      <p>A verification code will be sent to the email address entered.</p>
+      <p>Passwords must contain at least 8 characters, including an uppercase latter, a lowercase latter, a special character, and a number.</p>
+    </div>
+    <AmplifyAuthenticator :authConfig="authConfig" ref="authenticator" />
+  </secondary-layout>
+</template>
+
+<script>
+import { components } from 'aws-amplify-vue';
+import SecondaryLayout from '../components/SecondaryLayout/SecondaryLayout.vue';
+
+
+export default {
+  name: 'Auth',
+  components: {
+    AmplifyAuthenticator: components.Authenticator,
+    SecondaryLayout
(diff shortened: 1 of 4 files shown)


src/web-ui/src/public/Auth.vue
```diff
@@ -0,0 +1,166 @@
+<template>
+  <secondary-layout>
```
kazbaig, line 2: Nit pick: can we use `SecondaryLayout` for casing? I know it's not the `vue` way, but it's the naming convention we've followed thus far. Perhaps something to consider changing throughout the app.

src/web-ui/src/public/Auth.vue
```diff
@@ -0,0 +1,166 @@
...
+              type: 'string',
+              required: true
+            },
+          ]
+        }
+      }
+    }
+  },
+  mounted() {
+    this.$refs['authenticator'].$watch('displayMap', (newVal) => {
```
kazbaig, line 57: Another nit pick - dot notation would be ideal here. As it stands, might mess with IDE's renaming feature down the line if needed.

src/web-ui/src/public/Welcome.vue
```diff
@@ -32,7 +32,7 @@
           <div class="login-cta d-flex justify-content-center align-items-center">
             Have an account?<router-link to="/auth" class="sign-in btn btn-link">Sign in</router-link>
           </div>
-          <router-link to="/auth" class="create-account mt-3 mt-sm-0 ml-sm-3 btn btn-primary"
+          <router-link to="/auth?signup=true" class="create-account mt-3 mt-sm-0 ml-sm-3 btn btn-primary"
```
kazbaig, line 35: Since we've made the change the comment on line 29 is talking about, let's remove that comment as well.

kazbaig:
LGTM!

james-jory (approved):

PR #159, opened 2020-12-17
new products dataset release

pierreaws:
*Issue #, if available:*

*Description of changes:*

new products dataset from MLSL Demo Team.  The images bucket link is temporary 


stage.sh
```diff
@@ -86,7 +86,7 @@ done
 
 # Sync product images
 echo " + Copying product images"
-aws s3 sync ./images s3://${BUCKET}/${S3PATH}images --only-show-errors $S3PUBLIC
+aws s3 sync s3://rds-dataset-releases/1.3/images/  s3://${BUCKET}/${S3PATH}images/ $S3PUBLIC
```
alexchirayath, line 89: This would need to be updated before merging in.
@james-jory  please let us know a bucket where you would like us to copy all of the image data

alexchirayath (commented):
Just a comment pending on @james-jory  to provide us an s3 bucket to add the image. The rest of the PR looks good.
Please request a re-review once the bucket is updated in stage.sh

alexchirayath (changes requested):
1. The sizes of the screenshots added is too large (You can see in your commit, some images have increased by 400%) which lead to issues with CodeCommit upload. Please resize the images to be the same size(if not smaller)

2. The isntructions near emulate shopper profile need to be updated to reflect the new UX. Could you please make changes to it?

workshop/1-Personalization/1.1-Personalize.ipynb
```diff
@@ -1668,7 +1668,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "product_id = 22\n",
+    "product_id = '020a5afe-fb13-4499-a1fa-8594d326eaa0'\n",
```
james-jory, line 1671: FUTURE: I think we should change this to a randomly selected product from the items dataframe to make this code more future proof should the products.yaml file change again.

stage.sh
james-jory: I created the bucket `retail-demo-store-assets` and sync'd the images from `rds-dataset-releases` to the `/datasets/1.3/images/` folder.

So the following line in `stage.sh` should work.

```bash
aws s3 sync s3://retail-demo-store-assets/datasets/1.3/images/  s3://${BUCKET}/${S3PATH}images/ $S3PUBLIC
```

james-jory (commented):
Pending the bucket change and verifying that staging and deployment works from this bucket as expected, all looks good to me. Let me know when that has been verified. Thank you!

alexchirayath (approved):
LGTM. Thanks for the changes @pierreaws

stage.sh
james-jory: Changed bucket to an existing one (`retail-demo-store-code`) that is already on the allow list.

james-jory (approved):
The Search workshop was timing out indexing the new larger catalog in one call so had to add logic to break products into batches for bulk indexing. Works fine now.

alexchirayath:
> The Search workshop was timing out indexing the new larger catalog in one call so had to add logic to break products into batches for bulk indexing. Works fine now.

Ah interesting. Thanks for making adding the bulk index logic !

PR #161, opened 2020-12-21
Conditional demo guide badges

kazbaig:
*Issue #, if available:*

*Description of changes:*
Makes the demo guide badges dynamic based on the recipe returned by the section's request. Since this now has the same functionality as the inline recipe ARNs, those are removed. The same UI will still show if an experiment is active for now. 

Additionally, adds a demo guide badge on the category page.


src/web-ui/src/public/Live.vue
```diff
@@ -178,8 +179,9 @@ export default {
       return data;
     },
     async getRelatedProducts() {
-      this.explainRecommended = null;
+      this.experiemnt = null
```
james-jory, line 182: `experiemnt` typo

james-jory (changes requested):
One minor issue

src/web-ui/src/public/Live.vue
kazbaig: Fixed

james-jory (approved):
Looks good. Thanks.

PR #163, opened 2020-12-22
Delivery Alexa Skill With Amazon Location Service

AkramDweikat:
Alexa Skill With Amazon Location Service including:
   - Alexa skill for C-store.
   - Integration with retail demo store including carts, orders.
   - Authentication vis retail demo store Cognito user pool for Alexa user.
   - Integration with Amazon Location Service place indexes and geofence.
   - Order confirmation email.
   - Amazon Pay payment.
   - New products + images for c-store.
   - Personalize changes to filter between different c-store vs retail demo scenarios.
   - Personalize logic for creating sensible c-store data.
   - Deployment in 2 steps for Alexa+retail demo store stacks.
   
   This work was done by Dae.mn Team.
   Co-authored-by: Akram Dweikat akram@akramdweikat.com
   Co-authored-by: Akram Dweikat akram.dweikat@dae.mn
   Co-authored-by: Damien Jade Duff damien.duff@daemonsolutions.com
   Co-authored-by: Joe Major joe.major@dae.mn


src/web-ui/src/authenticated/Profile.vue
```diff
@@ -67,22 +73,27 @@
...
-            <button class="btn btn-primary" v-on:click="saveChanges"> <i class="fas fa-spinner fa-spin" v-if="saving"></i> Save Changes </button>
+            <button class="btn btn-primary" v-on:click="saveChanges"><i class="fas fa-spinner fa-spin"
+                                                                        v-if="saving"></i> Save Changes
+            </button>
           </div>
-
         </div>
+        <h4>Location Demo</h4>
+        <hr/>
+        <router-link type="button" class="btn btn-primary" :to="{name:'Collections'}">
```
kazbaig, line 84: The profile page doesn't exist anymore. You could add a link in `UserDropdown` instead.

src/web-ui/src/public/Notifications.vue
```diff
@@ -0,0 +1,154 @@
...
+                title: 'Collection available',
+                text: collectionText,
+                content: orderListDiv
+              });
+            })
+          }
+        }
+      }
+    },
+    getProductImageURL (product) {
```
kazbaig, line 121: There is a function that does this in the `util` dir

src/web-ui/src/public/components/AmplifyMap.vue
```diff
@@ -0,0 +1,259 @@
...
+        .then(this.loadMap)
+    },
+    async loadMap() {
+      if (!this.storePosition) {
+        return
+      }
+      this.map = new mapbox.Map({
+        attributionControl: false,
+        center: this.storePosition || [0, 0],
+        container: "mapContainer",
```
kazbaig, line 56: The API can also take an `HTMLElement` so using a `ref` would be more preferable here.

src/web-ui/src/public/components/AmplifyMap.vue
```diff
@@ -0,0 +1,259 @@
+<template>
+  <div id="mapContainer" style="height: 100%; width: 100%"></div>
```
kazbaig, line 2: Let's use class instead of style attribute here

james-jory:
It looks like this PR duplicates most/all of #162 and also includes changes from the already-merged IVS PR. Would it be possible to get a PR that just includes the Alexa specific changes?

damiendaemon:
> It looks like this PR duplicates most/all of #162 and also includes changes from the already-merged IVS PR. Would it be possible to get a PR that just includes the Alexa specific changes?

Undoubtedly some changes will be divorcable from PR #162 (the Location Services PR) but many changes build on the changes made from there - in particular, what comes to mind is Geofence event handling, Location Services stack deployment, Location Services map display, orders and cart service back-end changes and collections display. It may be possible to remove the offers Personalize campaign with a little effort (with or without changes to make multiple data-set groups work in the deployment).

damiendaemon:
Hi all,

This Alexa skill PR has been updated for master and responses to input, including:

 - Merge latest from master
   - User selections
 - Simplified the process by moving the Alexa back-end Lambda stack into main CloudFormation as substack.
 - Move to workshop.
 - Cleaned up documentation a lot.
 - Extended testing.

 To see what has been changed since the last delivery (bundles together changes from master/main):

      git diff c72636348840b36f1474f825d71958aa533b261e delivery-alexa

**Please see the attached tar file for files to add to the images dataset.**:

[datasets.tar.gz](https://github.com/aws-samples/retail-demo-store/files/7090078/datasets.tar.gz)


The approach taken to staging in this iteration has been to load from datasets 1.3 and subsequently 1.4.

src/aws-lambda/alexa-skill-lambda/alexa-skill-lambda.py
```diff
@@ -0,0 +1,1414 @@
...
+        handler_input: As passed into the Alexa skill Lambda handler.
+
+    Returns:
+        str: Basket ID
+    """
+    session_attr = handler_input.attributes_manager.session_attributes
+    if 'BasketId' in session_attr:
+        return session_attr['BasketId']
+    else:
+        return None
```
james-jory, line 523: Is there a reason why the Carts service is not being used here?

src/aws-lambda/alexa-skill-lambda/template.env
```diff
@@ -0,0 +1 @@
+COGNITO_DOMAIN=
```
james-jory, line 1: Should be in `.env.template` now.

src/aws-lambda/personalize-pre-create-campaigns/personalize-pre-create-campaigns.py
```diff
@@ -256,6 +259,20 @@
...
+    all_categories_url = service_url + '/categories/all'
+    logger.info(f"Getting full list of categories from URL {all_categories_url}")
+    response = urllib3.PoolManager().request('GET', all_categories_url, retries=False)
+    if response.status != 200:
+        errmsg = f"{response.status}: Products service not working - reason: {response.reason} response: {response.data}"
+        logger.error(errmsg)
+        raise Exception(errmsg)
+    else:
+        categories = json.loads(response.data)
+        return categories
```
james-jory, line 273: There could be a race condition here after the Products service is created in ECS and before the Products service is initially built and deployed. As long as this function is called well into the Personalize pre-create process, it should be okay. Filters should be kept towards the end. Maybe note this with a comment?

src/aws-lambda/personalize-pre-create-campaigns/personalize-pre-create-campaigns.py
```diff
@@ -99,7 +99,10 @@
 filters_config = [
      {'arn_param': 'retaildemostore-personalize-filter-purchased-arn',
       'filter_name': 'retaildemostore-filter-purchased-products',
-      'filter_expression': 'EXCLUDE itemId WHERE INTERACTIONS.event_type IN ("OrderCompleted")'},
+      'filter_expression': 'EXCLUDE itemId WHERE INTERACTIONS.event_type IN ("OrderCompleted") | EXCLUDE itemId WHERE ITEMS.CATEGORY NOT IN (%%%CATEGORY_LIST%%%)'},
+      {'arn_param': 'retaildemostore-personalize-filter-cstore-arn',
+       'filter_name': 'retaildemostore-filter-cstore-products',
+       'filter_expression': 'EXCLUDE itemId WHERE ITEMS.CATEGORY IN(%%%CATEGORY_LIST%%%)'}
```
james-jory, line 105: Using a hard-coded category list in the filter expression will make this brittle for catalogs with a large number of categories and catalogs where categories are added after deployment. Would using a product attribute be more scalable and performant? For example, for the purchase filter, could a product flag such as "REPEAT_PURCHASE" be used to not exclude products that are repeat purchased (i.e. all the c-store products) or maybe a "DEPARTMENT" product field that is "c-store" for all c-store products.

src/aws-lambda/retaildemostore-lambda-load-products/src/main.go
```diff
@@ -161,6 +162,8 @@ func loadData(s3bucket, s3file, ddbtable, datatype string) (string, error) {
 // HandleRequest - handles Lambda request
 func HandleRequest(ctx context.Context, event cfn.Event) (physicalResourceID string, data map[string]interface{}, err error) {
 
+    fmt.Sprintf("Received event: %s!", event )
```
james-jory, line 165: This does nothing with the resulting string. Remove or do something with the string.

src/recommendations/src/recommendations-service/app.py
```diff
@@ -159,12 +173,19 @@ def get_products(feature, user_id, current_item_id, num_results, campaign_arn_pa
...
         campaign_arn = values[0]
         filter_arn = values[1]
 
         if campaign_arn and (user_id or not user_reqd_for_campaign):
+
+            logger.info(f"get_products: Supplied campaign: {campaign_arn} (from {default_campaign_arn_param_name}) Supplied filter: {filter_arn} (from {default_filter_arn_param_name}) Supplied user: {user_id}")
+
+            if filter_arn is not None and (user_id is None or len(user_id.strip()) == 0):
+                logger.warning(f"userID is required if using filter - filter Arn: {filter_arn} - spoofing user to 0")
+                user_id = "0"
```
james-jory, line 187: UserId is always required for User-Personalization and Personalized-Ranking. For SIMS, it is only required when there is a filter **and** the filter references the users dataset. Since none of the filters in the RDS use the users dataset in their expression, a userId should not required.

workshop/5-Conversational/5.2-AlexaHandsfree.md
```diff
@@ -0,0 +1,515 @@
...
+#### Pre-Requisites
+
+These instructions assume that the Retail Demo Store has been deployed 
+ using CloudFormation in an AWS account for which you have CLI access configured. 
+ You will need to access the CloudFormation stack you created in the AWS Console UI
+ to be able to configure Alexa to work with the Retail Demo Store.
+
+#### Step-by-step
+
+The following instructions can be run from within a Sagemaker, Cloud9 or local terminal session.
```
james-jory, line 83: Minor issue: "Sagemaker" should be "SageMaker" (throughout this file).

james-jory (changes requested):

src/aws-lambda/alexa-skill-lambda/alexa-skill-lambda.py
damiendaemon: We have talked in our team and it is not clear. We cannot see how using the native carts service would not be fine to do also. Possibly it was simply considered expedient during development to keep the complexity down. I am hoping it is not a blocker for the current iteration?

src/aws-lambda/personalize-pre-create-campaigns/personalize-pre-create-campaigns.py
damiendaemon: Yes, the Personalize lambda will fail till products are loaded. We will add the comment.

src/aws-lambda/personalize-pre-create-campaigns/personalize-pre-create-campaigns.py
damiendaemon: I think the general option would be to have all of a "visible_ui", "visible_alexa" and "repeat_purchased" flags because there are these different concerns. Would you agree? Regarding the "after deployment" issue - changing to a product attribute would, yes, make the filter reflect the current state after the latest products list as uploaded to Personalize which though not currently done in Retail Demo Store in general makes sense. We will implement this and get back to you as quickly as we can.

damiendaemon: Or, to reduce the number of fields sent to Personalize and keep them naturally as string we can have WHERE_VISIBLE ('UI' or 'ALEXA' etc.) and RECOMMEND_PATTERN ('NO_REPEAT' or 'REPEAT' etc.)

damiendaemon: We have implemented the latter approach and are in testing.

damiendaemon: @james-jory Unfortunately the filter language does not allow us to include WHERE_VISIBLE=UI while excluding products that are in interactions with EVENT_TYPE=OrderComplete and have RECOMMEND_PATTERN=NO_REPEAT. This is because we cannot put the latter two conditions in a single WHERE clause (because they are from different datasets) and when chaining include and exclude statements we subtract the union of exclude statements from the union of include statements, so there is no way I can see to slice these two concerns while at the same time finding the set of items that are non-repeatable AND have OrderComplete interactions. So I am going to remove the RECOMMEND_PATTERN product attribute for now and it will be essentially hard-coded. The filters are going to look like :

damiendaemon:
@james-jory These suggestions of yours have been addressed. In particular:
     - We now filter based on product attribute rather than category list (means no need to load categories).
     - We fixed the issue of logs that do not print logs on products service.
     - We added defaults for EventEngine.
     - Removed unused template.env
     - We no longer check for user on filter
     - Sagemaker -> SageMaker
Hoping this is satisfactory.

src/aws-lambda/alexa-skill-lambda/alexa-skill-lambda.py
james-jory: Yes, I think this is a problem. This project is primarily an educational tool for customers to learn how to build similar experiences in their applications. Specifically, this breaks the notion of multi-modal user experiences where, say, I start putting items in a cart using the Alexa skill and then I want to complete my purchase using my phone or a kiosk. By having two cart implementations, continuity is lost. It adds confusion to customers as to why these choices were made and will make the project more difficult to maintain in the future.

src/aws-lambda/personalize-pre-create-campaigns/personalize-pre-create-campaigns.py
james-jory: So c-store products will not be visible or recommended in the web UI? If so, remind me why that is a requirement.

src/aws-lambda/personalize-pre-create-campaigns/personalize-pre-create-campaigns.py
```diff
@@ -256,6 +264,23 @@
                      'filename': offer_interactions_filename}}
 
 
+def get_all_product_categories():
```
james-jory, line 267: Is this still being used with the new filter approach?

james-jory (changes requested):
I think the basket/cart issue and the filters issue need to be thought through a bit more. Let's connect and discuss to find a path forward. Thanks!

src/aws-lambda/personalize-pre-create-campaigns/personalize-pre-create-campaigns.py
james-jory: > **Please see the attached tar file for files to add to the images dataset.**:
> 
> [datasets.tar.gz](https://github.com/aws-samples/retail-demo-store/files/7090078/datasets.tar.gz)
> 
> The approach taken to staging in this iteration has been to load from datasets 1.3 and subsequently 1.4.

The 1.4 directory has been recursively copied to the staging bucket for catalog images.

src/aws-lambda/personalize-pre-create-campaigns/personalize-pre-create-campaigns.py
damiendaemon: Deleted in the latest changes.

damiendaemon:
Latest changes:


Use carts service from retail demo store rather than a separate Alexa cart.
Use "guest" user when no Cognito user.
Require a profile user if Cognito account linked
Change invocation name as a workaround for stale skills in test Developer env
Remove "where_visible" flag due to lack of workarounds for filters.
Special purpose but more relevant Personalize filters.
Remove unused get_all_product_categories()
Add new categories to categories for upload
Merge from main

src/products/src/products-service/product.go
```diff
@@ -7,18 +7,18 @@ package main
...
+	SK               string  `json:"sk" yaml:"sk"`
+	Name             string  `json:"name" yaml:"name"`
+	Category         string  `json:"category" yaml:"category"`
+	Style            string  `json:"style" yaml:"style"`
+	Description      string  `json:"description" yaml:"description"`
+	Price            float32 `json:"price" yaml:"price"`
+	Image            string  `json:"image" yaml:"image"`
+	Featured         string  `json:"featured,omitempty" yaml:"featured,omitempty"`
+	GenderAffinity   string  `json:"gender_affinity,omitempty" yaml:"gender_affinity,omitempty"`
+	CurrentStock     int     `json:"current_stock" yaml:"current_stock"`
```
james-jory, line 21: Aliases is not in the field list but is loaded by the catalog load lambda. Was it left out on purpose?

james-jory (commented):
Looks good. Just one comment/question.

src/products/src/products-service/product.go
damiendaemon: Perhaps this dropped off while testing the product fields for the filters. An oversight, at any rate. I have added it back on.

james-jory (approved):
Looks good to me. Thank you!

PR #171, opened 2020-12-28
Recommendations as list

kazbaig:
*Issue #, if available:*

*Description of changes:*
Since users are unlikely to browse recommendations in a carousel we will be rendering the first recommendations section as a list of products instead of using a carousel. This will also make the home page appear more full.

Includes a refactor of the `ProductCarousel` to reuse the `Product` component.


james-jory (approved):
Nice changes.

PR #177, opened 2020-12-30
Fix Elasticsearch pre-index Lambda

james-jory:
*Issue #, if available:*

*Description of changes:*

The Elasticsearch pre-index Lambda (custom resource) was timing out indexing the new (larger) product catalog introduced in #159. 

- Increased function timeout from 1 minute to 5 minutes (testing was taking just over 1 minute so added buffer to support future expansion of catalog).
- Fix IAM role for function so that CW logs could be written. This will help with future debugging.
- Rather than indexing products one at a time, switched to bulk indexing to speed up overall process.


alexchirayath:
LGTM.

PR #190, opened 2021-01-08
Fix timeouts indexing products and optimize loading

james-jory:
*Description of changes:*

Resolving some recurring issues with deployments.

- When pre-indexing in Elasticsearch is selected, the lambda would sometimes timeout when the ES server would need to merge segments or the JVM would GC. The client now has a longer timeout and will retry. This should make this lambda more reliable.
- The custom resource lambda that loaded products in DDB was loading them for all CFN event types. Now it only loads on "Create" event types.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ src/aws-lambda/*/package
 src/aws-lambda/elasticsearch-pre-index/products.yaml
 src/aws-lambda/personalize-delete-resources/models
 src/aws-lambda/personalize-pre-create-campaigns/models
+src/aws-lambda/personalize-pre-create-campaigns/data
 src/aws-lambda/pinpoint-auto-workshop/pinpoint-templates
 local/*
 demo.md
@@ -26,3 +27,4 @@ generators/*.gz
 workshop/data/*
 workshop/datagenerator/*
 workshop/requirements.txt
+workshop/1-Personalization/generate_interactions_personalize.py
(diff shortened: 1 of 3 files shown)


alexchirayath:
LGTM. Has this been tested on an instance?
Also, one callout here is that if  the product list is updated, the stack update does not have any logic to update the product list

alexchirayath (approved):

james-jory:
> LGTM. Has this been tested on an instance?

Yes, it has been tested on a fresh deployment.

> Also, one callout here is that if the product list is updated, the stack update does not have any logic to update the product list

Yes. That is unrelated to these changes, though. My change to the product load lambda was really to prevent the lambda from trying to load the product and category DDB tables during a CFN "Delete" event. Besides being unnecessary, I was seeing some instability during CFN teardown so was making the logic more precise. Updating the products/categories YAML files in the staging bucket and updating the stack will not trigger an "Update" to the product load custom resource since those files are external to the CFN template. The custom resource itself would need to be updated to trigger the lambda to run with an "Update" event type. Nevertheless, I added the "Update" event to the conditional logic that loads the tables so it's there if we ever add something like a version template parameter that can be used to trigger the resource to update. With that said, you can still run the Lambda directly to update products from the YAML file (I've done this many times before).

manbearshark (commented):
LGTM!

manbearshark (commented):
Maybe the main.go fn could return an error when an unexpected query type is sent?

PR #196, opened 2021-01-19
update cart service and FE to use common REST API urls

svozza:
*Issue #, if available:*
#158

*Description of changes:*

As mentioned in the above issue, this pull requests makes the URLs in the cart service follow the standard pattern used by REST APIs. I have not changed the `cart/{id}` endpoint to return a 404 when no cart is found because it appears the frontend relies on the the object with a null `items` field returned when a cart is not found immediately after submitting an order. I will do the products service as soon as this PR is merged to minimise the amount of time where the APIs across the services are inconsistent.

I also updated the `go.mod` file because the container uses v1.15 of Go but the file format was for v1.14

### DIFF
--- a/src/carts/src/carts-service/routes.go
+++ b/src/carts/src/carts-service/routes.go
@@ -26,13 +26,13 @@ var routes = Routes{
 	Route{
 		"CartsIndex",
 		"GET",
-		"/carts/all",
+		"/carts",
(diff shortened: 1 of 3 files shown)


james-jory (approved):
Changes look good. Thanks!

PR #198, opened 2021-01-20
Imagesupdate

pierreaws:
*Issue #, if available:*

The documentation needed images and text updates.  

*Description of changes:*

The documentation was referring to old images that were converted since from png to jpg.  Once image was no longer existing, so it was removed here.  The user emulation text bit was updated to navigate the new layout.  Figures numbering was updated due to the removal of one image.  Also updated are the paragraphs describing the products categories,  from 8 to now 16, and their listing.


james-jory (approved):
Thanks!

PR #202, opened 2021-01-23
IVS integration fixes

### LINKED ISSUE #201: IVS product mappings are broken/missing
When the new products and categories were introduced in #159, the products added for the IVS integration (#96) were lost. This broke the [product mapping file](https://github.com/aws-samples/retail-demo-store/blob/master/videos/default_streams/default_streams.json) that maps products being demonstrated in the video stream with products being displayed and recommended on the live stream view.

The products added by the IVS PR need to be added back to the products.yaml file and the videos mapping file needs to be updated with the new product IDs.

james-jory:
*Issue #, if available:*

Closes #201 

*Description of changes:*

- Add IVS products back to the products.yaml file.
- Fix IVS video mapping to products to reflect new UUID for images added back to dataset.
- Upload IVS product images to the images staging bucket (not visible in this PR but done).
- Remove `images` folder from repo since it's no longer relevant.
- Add logic back to deal with missing product images (i.e. when a new product is added via the API but an image URL/path isn't set).


alexchirayath (changes requested):
Copy-pasting some of the feedback discussed outside of GitHub
 

products.yaml file seems to have formatting errors.   IVS products have the “-” character positioned with “category”.  As I understand, only the “id” should have it
 
- category: beauty  (wrong)
- id: ad41a106-99f7-4258-8060-d1dcb7de93b8  (correct)
 
Also,  some IVS products have invented a few styles that are somewhat redundant with the dataset.   The 3 IVS products below could be reclassified in the existing categories and style of our dataset.  If that’s not messing up something somewhere else,  then we might consider changing them this way.     
 
 Item                                   IVS                                                                          Dataset
80 (standing mirror)       housewares, bedroom        à     homedecor, mirror
81 (sofa)                            housewares,salon           à     furniture, sofas  
82 (cushion)               housewares,salon           à     homedecor, cushion

james-jory:
> Copy-pasting some of the feedback discussed outside of GitHub
> 
> products.yaml file seems to have formatting errors. IVS products have the “-” character positioned with “category”. As I understand, only the “id” should have it
> 
> * category: beauty  (wrong)
> * id: ad41a106-99f7-4258-8060-d1dcb7de93b8  (correct)

This is actually not an error. The [YAML spec](https://yaml.org/spec/1.2/spec.html#id2765608) does not enforce key order. The file was successfully loaded and products successfully added to DDB and ES when I tested the changes. Nevertheless, for convention with other products, I changed the file to move the "id" field to the top of the field list for the re-added products.

> Also, some IVS products have invented a few styles that are somewhat redundant with the dataset. The 3 IVS products below could be reclassified in the existing categories and style of our dataset. If that’s not messing up something somewhere else, then we might consider changing them this way.
> 
> Item IVS Dataset
> 80 (standing mirror) housewares, bedroom -> homedecor, mirror
> 81 (sofa) housewares,salon -> furniture, sofas
> 82 (cushion) housewares,salon -> homedecor, cushion

Good suggestion. Changed category and style mappings suggested above.

alexchirayath (approved):
LGTM

PR #209, opened 2021-02-22
Add support for x-ray for recommendations and search (python services)

BastLeblanc:
*Issue #, if available:* N/A

*Description of changes:* Add support for x-ray for 2 services:
- search
- recommendations

### DIFF
--- a/aws/cloudformation-templates/services/service/service.yaml
+++ b/aws/cloudformation-templates/services/service/service.yaml
@@ -266,7 +266,7 @@ Resources:
 
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/AWSCloudMapDiscoverInstanceAccess
-        - arn:aws:iam::aws:policy/AWSXrayFullAccess
+        - arn:aws:iam::aws:policy/AWSXRayDaemonWriteAccess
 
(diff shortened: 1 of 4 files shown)


aws/cloudformation-templates/services/service/service.yaml
```diff
@@ -266,6 +266,7 @@ Resources:
 
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/AWSCloudMapDiscoverInstanceAccess
+        - arn:aws:iam::aws:policy/AWSXrayFullAccess
```
james-jory, line 269: Is full access required or can this be scoped down?

james-jory (commented):
The `videos` service is also written in python. Will you instrument that service too?

What about the golang services?

aws/cloudformation-templates/services/service/service.yaml
BastLeblanc: scoped down to  arn:aws:iam::aws:policy/AWSXRayDaemonWriteAccess 
It's the best we can do (X-ray doesn't support resource level on this action)

james-jory (approved):

Issue #212, opened 2021-03-01
Videos microservice fails to deploy


blanco-lam:
Traceback (most recent call last):
File "app.py", line 341, in <module>
start_streams()
File "app.py", line 181, in start_streams
if is_ssm_parameter_set(SSM_VIDEO_CHANNEL_MAP_PARAM) and not USE_DEFAULT_IVS_STREAMS:

File "app.py", line 135, in is_ssm_parameter_set
response = ssm_client.get_parameter(Name=parameter_name)

File "/usr/local/lib/python3.8/site-packages/botocore/client.py", line 337, in _api_call
return self._make_api_call(operation_name, kwargs)

File "/usr/local/lib/python3.8/site-packages/aws_xray_sdk/ext/botocore/patch.py", line 38, in _xray_traced_botocore
return xray_recorder.record_subsegment(
File "/usr/local/lib/python3.8/site-packages/aws_xray_sdk/core/recorder.py", line 428, in record_subsegment
subsegment = self.begin_subsegment(name, namespace)
File "/usr/local/lib/python3.8/site-packages/aws_xray_sdk/core/recorder.py", line 292, in begin_subsegment
segment = self.current_segment()
File "/usr/local/lib/python3.8/site-packages/aws_xray_sdk/core/recorder.py", line 270, in current_segment
entity = self.get_trace_entity()
File "/usr/local/lib/python3.8/site-packages/aws_xray_sdk/core/recorder.py", line 390, in get_trace_entity
return self.context.get_trace_entity()
File "/usr/local/lib/python3.8/site-packages/aws_xray_sdk/core/context.py", line 99, in get_trace_entity
return self.handle_context_missing()
File "/usr/local/lib/python3.8/site-packages/aws_xray_sdk/core/context.py", line 124, in handle_context_missing
raise SegmentNotFoundException(MISSING_SEGMENT_MSG)
aws_xray_sdk.core.exceptions.exceptions.SegmentNotFoundException: cannot find the current segment/subsegment, please make sure you have a segment open

INFO in app: VIDEO_BUCKET: retail-demo-store-us-west-2
INFO in app: SSM_VIDEO_CHANNEL_MAP_PARAM: retaildemostore-ivs-video-channel-map
INFO in app: USE_DEFAULT_IVS_STREAMS: True
INFO in app: Starting video streams
cannot find the current segment/subsegment, please make sure you have a segment open

BastLeblanc:
hot fix merged. 
[...]

PR #213, opened 2021-03-02
hot-fix videos x-ray support

BastLeblanc:
*Issue #, if available:* #212 

*Description of changes:* AWS X-ray need to start a segment before any downstream call.
The video service is calling AWS service in the init phase, so forcing a `begin_segment` is required

### DIFF
--- a/src/videos/src/videos-service/app.py
+++ b/src/videos/src/videos-service/app.py
@@ -8,6 +8,8 @@ from aws_xray_sdk.core import patch_all
 
 patch_all()
 
+xray_recorder.begin_segment("Videos-init")
+
 import logging
 import json
 import os


src/videos/src/videos-service/app.py
```diff
@@ -8,6 +8,8 @@
 
 patch_all()
 
+xray_recorder.begin_segment("Videos-init")
```
james-jory, line 11: This will cover the default mode of using the default streams but I don't think it deals with the case where channels are provided and background threads are started for each channel. For that, I think we need to either annotate the `stream` function to begin/end segments or add code within the `stream` function for this.

james-jory (approved):
Approving for now as a hot-fix since it addresses the default deployment case but more work is needed so that it works when channels are specified.

PR #218, opened 2021-03-15
Web UI dependency updates

james-jory:
*Description of changes:*

- Bump hightlight.js and immer.js versions due to vulnerability reports.
- Local web-ui testing fix to set default Segment write key to `NONE` in .env file which is expected in the code and index.html for not-configured state.


src/web-ui/package.json
```diff
@@ -11,7 +11,7 @@
   "dependencies": {
     "@optimizely/optimizely-sdk": "4.0.0",
     "amplitude-js": "^5.8.0",
-    "aws-amplify": "^3.0.23",
+    "aws-amplify": "^3.0.24",
```
kazbaig, line 14: This on its own actually resolves all the dependency issues according `npm audit`. This is the case even with all of the dependencies removed.

src/web-ui/package.json
```diff
@@ -41,6 +41,8 @@
   "resolutions": {
     "acorn": "^6.4.1",
     "elliptic": "^6.5.3",
+    "highlight.js": ">=10.4.1",
```
BastLeblanc, line 44: do you need to force these ? is it not induced by amplify?
as suggested I would reduce to minimum versioning

james-jory: Yes, for highlight.js. The immer issue can be resolved by upgrading aws-amplify.

kazbaig:
Not seeing any vulnerabilities from `npm audit` after installing but I do see some changes in `package-lock.json` that need to be committed.

Also, I fresh installed my packages after removing all the packages in the `resolutions` field and still saw 0 vulnerabilities from `npm audit` so I still think the resolutions are not needed.

Issue #200, opened 2021-01-21
Focused/filtered catalog deployments


james-jory:
Add support for deploying and/or displaying products from a specific set of categories in the Retail Demo Store UX. This will allow demos to be delivered that are restricted to a type of products that are relevant to the person receiving the demo. For example, if delivering a demo to someone in the personal electronics space, restrict the deployment just to products in the appropriate categories. This could be implemented a number of ways from deployment or just in the UI at run-time (and with a Personalize dynamic filter).

aeggbeg:
This is a important ticket. Basically I want to create 4 demo's for 4 different industry e.g. Retail, Banking etc. It would be great to have a solution : 
1. apply some filter - similar to "don't show products that customer just bought" in Amazon Personalise
2. somehow with a query parameter I would like to say "industry=retail" and it shows the retail products. 

I am open to either option as I am now manually figuring out how to update/load dynamodb with new product image/product data. Benefit of option 2 is I can run same instance for multiple industry demo. In both cases we will need to update local customer specific product data/information - to make the demo more compelling to customers industry. Is there a way a short cut / easiest way to load product data to an existing deployment maybe. 

Just some thoughts based on my current use cases. Cheers

patrick-239:
Working on it, hope to create a PR this Friday.
@aeggbeg could you share your datasets for different demos?

PR #220, opened 2021-03-24
Focused catalog deployments - Issue #200

patrick-239:
*Issue #, if available:* 200

*Description of changes:*
This PR brings "fashion", "original" and "all" deployments types as well as manual how to add more types based on a current product dataset. We could also expand product catalog and add more categories and deployment types. 

Instead of using filters (it might introduce inconsistent recommendations for profile without specific category) I created a way how to deploy limited set of products. 

PS
Huge "thank you" for an UI developer as it automatically adapt to a changes in a backend.  


james-jory (commented):
I have some concerns with approach taken with this PR. I wish we could have discussed this before development was started.

This this approach, a lot of work is required each time someone wants to create a new combination of categories and will litter the source code with logic for each combination. It also more tightly binds the deployment to the current RDS catalog. This will make it harder to support bring-your-own-catalog in the future.

Instead can this be implemented once such that the user optionally provides the categories they want included at deployment and the deployment dynamically handles the rest? This will require moving the data generation step from staging to deployment (maybe as a conditional custom resource) but it will make the feature much more flexible.

Thoughts?

PR #234, opened 2021-05-19
Web UI deps

james-jory:
*Description of changes:*

Changes to resolve **all** pending and past dependency updates.

- Removed `resolutions` from package.json file for older/obsolete dependency overrides
- Ran `npm audit fix`
- Verified all past and pending dependabot alerts are resolved (I will remove those PRs once this one is approved and merged)
- Rebuilt and tested the webapp to check for any new issues caused by dependency updates


kazbaig:
A few points
- The `preinstall` script can be removed since we're not forcing resolutions anymore
- `moment` and `dotenv` can be used. Would need to remove a couple lines of code for `moment` but I don't think those lines are achieving anything
- Running `npm install` is currently updating `package-lock.json` for me. This should updated version should be committed (I know there are updates to that file as part of this PR so not sure what's going on there)
- There are still some vulnerabilities coming from vue CLI but they don't seem to be addressed yet. I wouldn't worry about forcing resolutions too much on dev dependencies though as they don't get shipped in your bundle; better to just wait for an official fix

james-jory:
> * The `preinstall` script can be removed since we're not forcing resolutions anymore

Good catch. Will remove.

> * `moment` and `dotenv` can be used. Would need to remove a couple lines of code for `moment` but I don't think those lines are achieving anything

Will remove moment but it looks like we use dotenv for process.env.

> * Running `npm install` is currently updating `package-lock.json` for me. This should updated version should be committed (I know there are updates to that file as part of this PR so not sure what's going on there)

Ran `npm install`, built, and tested the app again. Uploaded updated `package-lock.json`.

> * There are still some vulnerabilities coming from vue CLI but they don't seem to be addressed yet. I wouldn't worry about forcing resolutions too much on dev dependencies though as they don't get shipped in your bundle; better to just wait for an official fix

Sounds good.

Thanks!

kazbaig:
> looks like we use dotenv for process.env

[Vue CLI loads environment files automatically](https://cli.vuejs.org/guide/mode-and-env.html#modes)

james-jory:
> > looks like we use dotenv for process.env
> 
> [Vue CLI loads environment files automatically](https://cli.vuejs.org/guide/mode-and-env.html#modes)

Good catch. Removed dotenv, ran `npm audit fix` again (to pick up some additional dependabot alerts), and retested the webpp.

PR #246, opened 2021-07-15
Make Segment resources optional

james-jory:
*Description of changes:*

Added CloudFormation template parameter to conditionally deploy Segment resources (Lambda functions and associated roles) so that internal AWS deployments can skip Segment and avoid role audit alerts.


manbearshark (approved):
LGTM!

PR #247, opened 2021-07-23
Go proxy override

james-jory:
*Description of changes:*

When developing with the products service locally and your network policy blocks access to the default Go module proxy, it's difficult to get the products service up and running in local Docker instance without having to manually hack the Dockerfile. These changes allow the Go proxy to be easily overridden using a build argument that is set in your .env file by setting `GOPROXY_OVERRIDE` build argument to `direct`.


BastLeblanc:
looks good but why only for the products service ?

james-jory:
> looks good but why only for the products service ?

That is the only golang service that is currently using modules. When the others are switched to modules, they will need the same treatment.

PR #248, opened 2021-07-29
When we get an internal server exception in the IVS streaming server …

damiendaemon:
When we get an internal server exception in the IVS streaming server we want the
IVS streaming server to keep running.

 -----


james-jory (approved):
LGTM!

PR #249, opened 2021-07-29
Workaround for 'Address' lacking in Pinpoint Lambda

damiendaemon:
Sometimes Pinpoint does not supply endpoint addresses to its enrichment Lambda. If this happens, the messages do not send. We fill it back in if this happens.

-----


james-jory (approved):
Good catch. Thanks!

PR #251, opened 2021-08-04
Allow SMS testing with numbers that are not PhoneTypeCode 0

damiendaemon:
Not every number that can receive an SMS has PhoneTypeCode 0 - for example, toll-free numbers that can be obtained through Pinpoint and used for 2-way messaging.

-----

### DIFF
--- a/src/users/src/users-service/handlers.go
+++ b/src/users/src/users-service/handlers.go
@@ -370,45 +370,42 @@ func UserVerifyAndUpdatePhone(w http.ResponseWriter, r *http.Request){
 			fmt.Println(res)
 			mobilePhoneCode := int(*res.NumberValidateResponse.PhoneTypeCode)
 			if (mobilePhoneCode != 0) {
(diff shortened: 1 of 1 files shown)


damiendaemon:
The diff without whitespace is more informative:
https://github.com/aws-samples/retail-demo-store/pull/251/files?w=1

james-jory (commented):
Rather than just logging a message and allowing all phone types through, why not at least prevent known invalid phone types from getting through. Do you know what the phone type is for the Pinpoint 2-way number? Maybe "OTHER"? We should be able to at least deny LANDLINE and INVALID numbers.

https://docs.aws.amazon.com/pinpoint/latest/apireference/phone-number-validate.html#phone-number-validate-prop-numbervalidateresponse-phonetypecode

damiendaemon:
> Do you know what the phone type is for the Pinpoint 2-way number? Maybe "OTHER"? We should be able to at least deny LANDLINE and INVALID numbers.

It's VOIP in my experience, also for services like https://receive-smss.com/

I have updated the PR.

Note that the error message does not make it through to the UI at present.

Have a good day!

james-jory (approved):
Looks good. Thanks. 

Noted that we need to fix getting error messages like this back from the service and displayed to the user so they're more helpful.

PR #261, opened 2021-08-18
Move .env to template.env

damiendaemon:
This is a convenience for developers. When developing and switching between branches it becomes easy to accidentally check in `.env` since there is a template `.env` file checked in.
Here we ignore `.env` and instead check in a file called `template.env`.
We do this for the **docker-compose** `.env` and for the **web-ui**.

-----


james-jory (changes requested):
The current approach and the proposed change each have their pros and cons. The .env files are intended to provide suitable defaults that you can run with without having to copy them from a template file. With this change you'd have to copy the template before running locally but it does provide more protection against the unintended commit.

I'm okay with switching to this approach but would ask that the template file name be changed to `.env.template` so it will be more easily found by a developer looking for a `.env` file when it doesn't exist to start with.

damiendaemon:
`template.env` is now `.env.template`. In retrospect perhaps this would have been better as `.template.env` so that editors can pick the file type from the file extension. Happy to make that change if requested.

src/web-ui/README.md
```diff
@@ -6,14 +6,14 @@ When deployed to AWS, CodePipeline is used to build and deploy the Retail Demo S
...
 3. Under Pipelines, find the pipeline with "WebUIPipeline" in the name and click on its name.
 4. You should now see the stages for the "WebUIPipeline" listed: Source and Build. For the "Build" stage, click on the "Details" link in the "Build" box.
 5. In the "Build logs" tab, scroll down through the build log output until you see the output for the `cat .env` command. You should see a block of several lines that start with `VUE_APP_...`. These lines represent the environment variables used when building the Web UI assets when deployed to S3. Select all of these lines (except the variables with `***` as their values--these are resolved as system parameters) with your mouse and copy them to your clipboard. Alternatively, if you only want to connect to some Retail Demo Store services in ECS and others running locally, just copy the lines that you need for remote services.
-6. Open your local `.env` file and paste over the top of the lines you selected.
+6. Open your local `.env` file and paste over the top of the lines you selected. Tthere is already a [.env.template](.env.template) file with examples of what this would look like.
```
james-jory, line 16: Typo. "Tthere" s/b "There".

james-jory (changes requested):
Minor typo. Otherwise looks good.

src/web-ui/README.md
damiendaemon: The typo is fixed. Thanks!

james-jory (approved):
Looks good. Thanks!

PR #269, opened 2021-09-15
Product multi-item fetch optimization

### LINKED ISSUE #101: Add multi-product get API to the products service
The recommendations service APIs and search feature in the web-ui (SearchItem.vue) have logic where a list of product IDs is iterated over to retrieve details for each product as part of returning results or rendering a component. This results in multiple calls to the `/products/id/{id}` API (one call for each product). The performance for these features could be optimized if an alternative API on the products service allowed multiple product IDs to be specified in a single request and have details for all of those products returned in the response.

For example, 

`/products?id=1,2,3,4`

would return a list of products that includes details for products with the `id` of 1, 2, 3, or 4.

The DynamoDB [BatchGetItem](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html) API can be used to efficiently query items for this new API.

Impacted APIs in the recommendations service include:

`/recommendations`
`/related`

james-jory:
*Issue #, if available:*

Closes #101 

*Description of changes:*

- Added ability to request details for multiple products, based on product ID, in a single request to the Products service.
- Refactored the products DDB primary key and indexes so that the DDB BatchGetItem API can be used to efficiently request a batch of items.
- Updated Recommendations service to use the new multi-product GET API to fetch product details in a single call rather than separate calls for each item (optimization).
- Updated search component in the web app to use the new multi-product GET API to fetch product details in a single call rather than separate calls for each item (optimization).
- Added a utility script that can be used to more easily load and reload categories and products in local DDB or AWS DDB from local machine.

### DIFF
--- a/src/web-ui/src/partials/Navigation/Search/Search.vue
+++ b/src/web-ui/src/partials/Navigation/Search/Search.vue
@@ -27,7 +27,7 @@
       <SearchItem
         v-for="result in results"
         :key="result.itemId"
-        :product_id="result.itemId"
+        :product="result.product"
         :experiment="result.experiment"
         :feature="feature"
       />
@@ -46,6 +46,7 @@ import LoadingFallback from '@/components/LoadingFallback/LoadingFallback';
 
 const SearchRepository = RepositoryFactory.get('search');
 const RecommendationsRepository = RepositoryFactory.get('recommendations');
+const ProductsRepository = RepositoryFactory.get('products')
 
 const EXPERIMENT_FEATURE = 'search_results';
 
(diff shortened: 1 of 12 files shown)


src/web-ui/src/partials/Navigation/Search/Search.vue
```diff
@@ -101,6 +105,18 @@ export default {
         this.results = items;
       }
     },
+    async lookupProducts(items) {
```
kazbaig, line 108: Same thing but easier to parse:

```js
const itemIds = items.map(item => item.itemId);

const { data } = await ProductsRepository.getProduct(itemIds);

this.results = items.map((item) => ({
  ...item,
  product: data.find(({id}) => id === item.itemId)
}));
```

james-jory: Good suggestion. Made this change.

BastLeblanc (commented):
Happy with the new API arguments (multiple product ID).

Adding a GSI for category is good too. Can we get rid of the categories table now?

kazbaig (approved):

BastLeblanc (approved):
Approved!

PR #275, opened 2021-11-08
Do not regenerate product ID if supplied in new record endpoint

damiendaemon:
External applications may want to supply a product ID via the endpoint.
If supplied, we do not overwrite this ID.
Note: duplicates IDs will update the existing record (no checking for duplicates is done).
This change has been tested.

-----
Damien.Duff@dae.mn Joe.Major@dae.mn


BastLeblanc (approved):
Approved. 
It's by design not checking for duplicate.
If ID is specified : insert or update if existing
if no ID is specified: generate GUUID and insert

PR #276, opened 2021-11-09
mParticle Integration - Final Updates

manbearshark:
*Issue #, if available:*

*Description of changes:*

Updated the Retail Demo Store to include mParticle;  this is the second PR which incorporates James' comments and suggestions from the first PR.


aws/cloudformation-templates/mparticle.yaml
```diff
@@ -0,0 +1,135 @@
...
+        S3Key: !Sub '${ResourceBucketRelativePath}aws-lambda/mparticle-personalize.zip'
+      Runtime: nodejs12.x
+      Timeout: 900
+      FunctionName: !Sub '${Uid}-mParticlePersonalizeLambda'
+      Environment:
+        Variables:
+          PERSONALISE_TRACKING_ID:  ''
+          PERSONALISE_CAMPAIGN_ARN: ''
+          MPARTICLE_S2S_API_KEY: ''
+          MPARTICLE_S2S_SECRET_KEY: ''
```
james-jory, line 58: Shouldn't these be looked up from SSM?

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
```diff
@@ -0,0 +1,177 @@
...
+        const payload = JSON.parse(payloadString);
+        const events = payload.events;
+        var amazonPersonalizeUserId;
+        
+        // First, get the mParticle user ID from the events payload.  In this example, mParticle will send all the events
+        // for a particular user in a batch to this lambda.
+        // retreive the mParticle user id which is available for anonymous and known customer profiles
+        var anonymousID = events[0].data.custom_attributes.mpid.toString();
+        
+        // if the customer profile is known then replace the amazon Personalize User id with the actual
```
james-jory, line 58: Nit: Capitalize "amazon".

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
```diff
@@ -0,0 +1,177 @@
...
+        // personalize Id captured from the user's profile
+        if(payload.user_attributes && payload.user_attributes.amazonPersonalizeId)
+            amazonPersonalizeUserId = payload.user_attributes.amazonPersonalizeId; 
+        else
+            amazonPersonalizeUserId = anonymousID;
```
james-jory, line 63: For PutEvents calls for anonymous users, you actually want to exclude userId from the call. Later when the user does identify themselves and events are sent in with the actual userId, Personalize will associate the prior anonymous events with that user. By using the anonymousID here, you will block that from happening and Personalize will consider the events for the anonymous and known user to be for different users.

However, when calling GetRecommendations for an anonymous user, use the sessionId as the userId in the GetRecs call. This will allow you to get personalized recs for anonymous users before the next retraining where the association described above is done.

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
```diff
@@ -0,0 +1,177 @@
...
+        if(payload.user_identities){
+                for (const identityRecord of payload.user_identities)
+                {
+                    if(identityRecord.identity_type==="customer_id")
+                        customerId = identityRecord.identity; 
+                }
+            }
+
+        var params = {
+            sessionId: payload.message_id,
```
james-jory, line 76: Is `payload.message_id` different for each event or is it really a sessionId that is the same for all events for the user's session?

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
```diff
@@ -0,0 +1,177 @@
...
+                const action = e.data.product_action.action;
+                const event_id = e.data.event_id;
+             
+
+                let params = {
+                    sessionId: payload.message_id,
+                    userId: amazonPersonalizeUserId,
+                    trackingId: personalizeTrackerID,
+                    eventList: []
+                };
```
james-jory, line 94: Already done above on L75. Delete?

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
```diff
@@ -0,0 +1,177 @@
...
+
+                // Build the list of events for the user session...
+                for (const product of e.data.product_action.products) {
+                    const purchasedItem = { itemId: product.id };
```
james-jory, line 98: The `discount` column is missing. Should be a `Yes`/`No` value indicating if the product interacted with is discounted or not. Suggest defaulting to `No` if a value cannot be passed through mParticle.

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
```diff
@@ -0,0 +1,177 @@
...
+            }
+
+        var params = {
+            sessionId: payload.message_id,
+            userId: amazonPersonalizeUserId,
+            trackingId: personalizeTrackerID,
+            eventList: []
+        };
+
+        for (const e of events) {
```
james-jory, line 82: Will the number of events always be <= 10?

james-jory (changes requested):

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
hawjefferson: Hi James!

mParticle has the ability to do identity resolution from mParticle's end. Technically, every time a profile is created in mParticle, mParticle will associate the said user with a unique mParticle ID hence the user being tracked is semi-known to our system. Once the user decides to identify themselves within mParticle, we will actually retain the same mParticle ID but just append additional identifiers (e.g.: userId, email, etc..) to the same profile such that any previous information we tracked for the said user will be retained and moving forward, the identified profile will be the profile which will always be associated back to the same user. This is the reason why I have the else clause because that's where I'm linking the MPID we maintain for the said user being tracked my mParticle. Hope this information helps

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
hawjefferson: Noted! Will changed to "Amazon"

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
hawjefferson: The payload.message_id will always be unique for each event that gets sent to this lambda function. This is actually event_id from our end and every event we capture will always have a unique_id.

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
hawjefferson: Good catch! Deleted

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
hawjefferson: added discount Column to set to No

src/aws-lambda/mparticle-personalize/mparticle-personalize.js
hawjefferson: yes, the number will always be less than 10.

aws/cloudformation-templates/mparticle.yaml
hawjefferson: This won't be available during the creation of the stack within CF.  What Igor and I have done is basically, within the Jupyter notebook for mParticle, to set the auto-generated values after the stack has been created into SSM which will be pulled eventually by the dependencies. I'll let Igor explain further here as he was the one who helped me build the cloud formation template. Cheers

james-jory: If these values are pulled from SSM, are the environment variables needed here? The function does not do anything with these variables. It looks like these can be deleted.

james-jory:
Just one issue left to resolve... the environment variables declared in the CFN template for the mParticle function that appear to not be used. Please delete from CFN template if not needed.

aws/cloudformation-templates/mparticle.yaml
manbearshark: Fixed this issue - these are not needed as they are being set exclusively via SSM - we should move to using this model to simplify the CF templates as there is a lot of error-prone variable passing required otherwise

james-jory (approved):
Looks great! Thank you!

Issue #15, opened 2020-04-14
Update Carts service to use DynamoDB


james-jory:
Updates to carts are currently not being persisted to a durable datastore. This task would add DDB as the datastore back-end for carts. The products service can be used as a model.

PR #277, opened 2021-11-09
Added service URLs as outputs of deployment template (#15)

damiendaemon:
Bring the service domain names to the top level of cloud formation.
This is useful if:
 - we want to make use of the Retail Demo Store CloudFormation as a substack of another solution.
 - we frequently go looking for these URLs for the purposes of testing.

-----
Joe.Major@dae.mn


james-jory (approved):
LGTM

PR #286, opened 2021-12-01
feat: add Swagger UI service in docker-compose

ijemmy:
*Issue #, if available:*

*Description of changes:*
Add Swagger UI service in docker-compose.  Contributors/users can run this service locally and try executing service from Swagger UI page. (similar to [this example](https://petstore.swagger.io/))

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -47,7 +47,7 @@ When deployed to AWS, the Web UI is hosted in an S3 bucket and served by CloudFr
 
 ## Swagger UI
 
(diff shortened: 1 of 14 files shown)


BastLeblanc:
it looks good to me. you can add the other services before we merge?

Is it only deployed during local dev or also during a live instance? I think it would be good to deploy at the same time as the full retail demo store. We envision this swagger ui useful when other projects want to integrate to an existing instance of the Retail Demo Store.

BastLeblanc (approved):
Approved, looks good to merge.

PR #297, opened 2021-12-16
Change notebook instance to ml.t2.medium

### LINKED ISSUE #283: Notebook instance type ml.t3.medium is not available in Sydney region
We have done several deployments in Sydney region (ap-southeast-2) in different accounts and none of them have worked with notebook instance type ml.t3.medium (ml.t2.medium **does** work). Perhaps this region should not be listed as supported anymore?

james-jory:
*Issue #, if available:*
Closes #283 

*Description of changes:*

Since the ml.t3.medium SageMaker instance type is not supported across all of the regions and AZs that this project deploys to, switch back to ml.t2.medium which is more broadly supported.


BastLeblanc (approved):
No worries for the change. All good

Issue #309, opened 2022-02-21
Python service deployments failing due to Flask dependency issue


daemon-joe:
Deployments of all Flask services (Search, Location, Offers, Recommendations, Videos) are failing with the following error: 
```
Traceback (most recent call last):
  File "app.py", line 6, in <module>
    from aws_xray_sdk.ext.flask.middleware import XRayMiddleware
  File "/usr/local/lib/python3.8/site-packages/aws_xray_sdk/ext/flask/middleware.py", line 1, in <module>
    import flask.templating
  File "/usr/local/lib/python3.8/site-packages/flask/__init__.py", line 21, in <module>
    from .app import Flask, Request, Response
  File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 25, in <module>
    from . import cli, json
  File "/usr/local/lib/python3.8/site-packages/flask/json/__init__.py", line 21, in <module>
    from itsdangerous import json as _json
ImportError: cannot import name 'json' from 'itsdangerous' (/usr/local/lib/python3.8/site-packages/itsdangerous/__init__.py)
```

This is an occurrence of [this Flask issue.](https://github.com/pallets/flask/issues/4455) The issue has been recognised by the Flask team. 

Suggested fix: 
Immediate: pin the following dependencies: 
```
itsdangerous==1.1.0
markupsafe==1.1.1
```

Longer-term: upgrade Flask to a supported version, as 1.0.2 is no longer supported.

BastLeblanc:
thanks for the fix, PR is merged

PR #310, opened 2022-02-21
Bugfix: Pin Flask dependencies to avoid deployment failure

daemon-joe:
*Issue #, if available:*
#309 

*Description of changes:*
Specify pinned versions of Flask dependencies to prevent dependency issues:
```
itsdangerous==1.1.0
markupsafe==1.1.1
```


BastLeblanc (approved):
Confirmed [here](https://github.com/pallets/flask/issues/4455) that for Flask 1.x we need to pin the dependencies

PR #313, opened 2022-02-23
Bugfix: Location services deletion (#20)

daemon-joe:
*Issue #, if available:*
N/A

*Description of changes:*
Currently, the custom resource which provisions and manages Amazon Location Service resources through CloudFormation fails to properly delete resources, causing stack deletion to fail on the first attempt. 
- Return resource name from Location Service custom resource to ensure deletion completes as expected. 


james-jory (approved):
LGTM. Thank you.

PR #333, opened 2022-03-14
feat: add new flags to stage.sh to avoid public access s3 objects and rebuild all packages.

ijemmy:
*Issue #, if available:* N/A

*Description of changes:*
When working locally, we don't want to make s3 upload public or repackage all code every time. Currently, we have to comment some lines out during development and may accidentally commit and push it. This change adds a script to accept new flags to change the behaviour to suit local development.

This change adds a script to parse two boolean flags `--private-s3` and  `--only-cfn-template` for those two use cases.

This change is **Backward-compatible**.  CodeBuild currently run script by passing only one argument (the bucket). If you don't pass any flag, the script will work the same as it was.
image


stage.sh
```diff
@@ -41,52 +111,53 @@ echo " + Uploading CloudFormation Templates"
...
-for function in ./src/aws-lambda/*/
-do
-    echo "  + Staging $function"
-    cd $function
-    chmod +x ./stage.sh
-    ./stage.sh ${BUCKET} ${S3PATH} > ../../../local/stage.log
-    cd -
-done
-
+if [ "$only_cfn_template" = false ]; then
```
ijemmy, line 114: Note for reviewer: I only add `if` and `fi` and indent all the codes. I didn't change any logic here.

BastLeblanc (approved):
good, and should fix the regression introduced in the #321 (which had a commented line on public files in stage.sh)

PR #335, opened 2022-03-19
Ensure timestamp gets incremented

### LINKED ISSUE #334: Personalize interactions sometimes have the same user, item, event type, and timestamp
Analysis of the generated interactions dataset indicates that there are several instances of duplicate interactions with the same user ID, item ID, event type, and timestamp. After investigating the generator script, there are cases where the timestamp is not being incremented as it should be.

james-jory:
*Issue #, if available:*

Closes #334 

*Description of changes:*

- Move the logic to bump the next timestamp inside the for loop that generates multiple interactions for the same item.
- Ensure that at least one second is added to timestamp when interactions for multiple event types are added.


BastLeblanc (approved):
all good

PR #338, opened 2022-03-23
Products API integration tests

dairiley:
_Issue #, if available:_ 
N/A

_Description of changes:_
Adds integration tests for the Products API. The tests for the DELETE /products/id/{productID} and PUT /products/id/{productID}/inventory endpoints are currently disabled due to preexisting bugs


src/products/README.md
```diff
@@ -40,3 +40,39 @@ Updates the current inventory value for the product identified by `{productID}`.
 Returns details on all categories.
 ### GET /categories/id/{categoryID}
 Returns details on the category identified by `{categoryID}`.
+
+## Testing
+To run integration tests for the Products service a Python virtual environment is required. You must have Python 3.3+ installed on your system to run the commands below. The commands are written to be ran from the project root.
```
ijemmy, line 45:
```suggestion
To run integration tests for the Products service a Python virtual environment is required. You must have Python 3.8+ installed on your system to run the commands below. The commands are written to be ran from the project root.
```

Let's use 3.8.  Most of the Lambda runtime use 3.8 (some use 3.7) so let's keep it consistent.

src/products/README.md
```diff
@@ -40,3 +40,39 @@ Updates the current inventory value for the product identified by `{productID}`.
...
+
+The following command will create a virtual environment in the test directory of the products service. 
+```console
+python3 -m venv src/products/test/.venv
```
ijemmy, line 49: (nit) May be we tell users upfront that we will run all of this in `src/products/test` so that we don't need to put prefix for every line?

src/products/test/integ/json_schemas.json
```diff
@@ -0,0 +1,125 @@
+{
+  "/products/all": {
+    "type": "array",
+    "data": {
```
ijemmy, line 4: I believe it should be `items` here?
Could you double check if the "jsonschema.validate()" will fail if you modify  the schema?

src/products/test/integ/json_schemas.json
```diff
@@ -0,0 +1,125 @@
...
+      "type": "dict",
```
ijemmy, line 5: I think the type should be `"object"` instead of `"dict"`?

src/products/test/integ/json_schemas.json
```diff
@@ -0,0 +1,125 @@
...
+        "aliases": {"type": "null"},
+        "price": {"type": "number"},
+        "image": {"type": "string"},
+        "featured": {"type": "string"},
+        "current_stock": {"type": "number"}
+    }
+  },
+    "/products/featured": {
+    "type": "array",
+    "data": {
```
ijemmy, line 42: same here. Could you double check if should be "items" instead of "data"?

src/products/test/integ/json_schemas.json
```diff
@@ -0,0 +1,125 @@
...
+          "image": {"type": "string"},
+          "featured": {"type": "string"},
+          "gender_affinity": {"type": "string"},
+          "current_stock": {"type": "number"}
+      }
+    }
+  },
+  "/products/category/:category_name": {
+    "type": "array",
+    "data": {
```
ijemmy, line 63: same here.

src/products/test/integ/json_schemas.json
```diff
@@ -0,0 +1,125 @@
...
+          "aliases": {"type": "null"},
+          "price": {"type": "number"},
+          "image": {"type": "string"},
+          "current_stock": {"type": "number"}
+      }
+    }
+  },
+  "/categories/all": {
+    "type": "array",
+    "data": {
```
ijemmy, line 83: same here.

src/products/test/integ/test-products.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def assertions(r, endpoint):
+    """
+    Assert response is successful and validate response body when applicable.
+    """
+    assert r.status_code is 200
+    if r.headers["Content-Type"] == "application/json; charset=UTF-8":
+        assert validate_schema(r.text, endpoint) is True
+
```
ijemmy, line 77:
```suggestion
    if r.headers["Content-Type"].startswith("application/json"):
        assert validate_schema(r.text, endpoint) is True
    else:
        assert False

```

ijemmy (changes requested):

src/products/test/integ/test-products.py
ijemmy: Some API returns only "application/json". Change it to `startswith` will make this more safe.

And we can automatically return False here as we don't expect any other content type.

src/products/test/integ/test-products.py
```diff
@@ -0,0 +1,143 @@
...
+    Validate a JSON response body against a schema.
+    """
+    with open(absolute_file_path('json_schemas.json')) as f:
+        schema = json.loads(f.read())[endpoint]
+
+    try:
+        jsonschema.validate(json.loads(json_str), schema)
+    except jsonschema.exceptions.ValidationError as e:
+        return False
+    return True
```
ijemmy, line 90:
```suggestion
jsonschema.validate(json.loads(json_str), schema)
return True
```

We swallow the exception here.  This exception contains a very useful info for debugging (why the validation fails).  I think we could just leave it without `try-except`.

This is an error with the current code.
image

This is an error without try-except
image

src/products/README.md
```diff
@@ -40,3 +40,39 @@ Updates the current inventory value for the product identified by `{productID}`.
 Returns details on all categories.
 ### GET /categories/id/{categoryID}
 Returns details on the category identified by `{categoryID}`.
+
+## Testing
+To run integration tests for the Products service a Python virtual environment is required. You must have Python 3.8+ installed on your system to run the commands below. The commands are written to be ran from the test directory of the products service (`src/products/test`).
+
+The following command will create a virtual environment.
```
BastLeblanc, line 47: do we have to have a virtual env for each services?

src/products/README.md
```diff
@@ -40,3 +40,39 @@ Updates the current inventory value for the product identified by `{productID}`.
...
+## Testing
+To run integration tests for the Products service a Python virtual environment is required. You must have Python 3.8+ installed on your system to run the commands below. The commands are written to be ran from the test directory of the products service (`src/products/test`).
+
+The following command will create a virtual environment. 
+```console
+python3 -m venv .venv
+```
+
+Some environment variables are required to run the tests and need to be added to the virtual environment. The example below will work for local development. Change as required depending on environment.
+```console
```
BastLeblanc, line 53: reference the new CloudFormation output maybe ?

src/products/test/integ/test-products.py
```diff
@@ -0,0 +1,142 @@
...
+products_api_url = os.getenv('PRODUCTS_API_URL')
+cwd = os.path.dirname(os.path.abspath(__file__))
+wildcards = {
+    ":product_id": os.getenv('TEST_PRODUCT_ID'),
+    ":category_name": os.getenv('TEST_CATEGORY_ID'),
+    ":category_id": os.getenv('TEST_CATEGORY_ID')
+}
+
+
+def full_request_url(text):
```
BastLeblanc, line 16: these functions are very generic, have a library somewhere instead of in each service?

src/products/README.md
dairiley: Can you clarify which part of the output should be referenced here? As the tests are separate I’m not sure it's relevant.

src/products/README.md
dairiley: Yes, it's being done in order to keep the tests as standalone services.

src/products/README.md
BastLeblanc: referring to


```sh
aws cloudformation describe-stacks --stack-name retaildemostore \
  --region REGION \
  --query "Stacks[0].Outputs[?OutputKey=='ExportEnvVarScript'].OutputValue" 
  --output text
```

in src/README.md

src/products/test/integ/test-products.py
dairiley: I've moved these methods to their own module so we can reuse them for the other services.

src/products/README.md
dairiley: I think the only relevant variables there are the service URLs for the deployment so I believe we can leave this out or just point the user in the correct direction of the service URLs only rather than showing everything?

BastLeblanc (approved):

PR #346, opened 2022-04-05
Fenix demo

ronaks4:
*Issue #, if available:*
https://app.asana.com/0/1200678426853792/1202019110694179/f
*Description of changes:*
This issue includes FenixCommerce Smart delivery dates on Product page, Category Page, Cart Page, Checkout Page.
It is front-end integration. UI will call Fenix API to get data on Estimated Delivery Date (EDD)
    Integration uses 3rd Party API https://ipinfo.io/ to detect zipcode.
    Based on product ids and SKU we see product delivery dates estimates on different sections.
This is phase-1 with on/off via deployment template. All products from demo store is pre-loaded in Fenix platform and rules for delivery estimate are per-configured.

Here are few sample screenshots the UI look.
Product Page:

image

Deployed full RDS in my account for multiple times and testing by Partner team and PSA.


aws/cloudformation-templates/base/_template.yaml
```diff
@@ -102,6 +102,52 @@ Parameters:
     Type: String
     Description: The GitHub branch used for deployment.
 
+  FenixZipDetectUrl:
+    Type: String
+    Description: Fenix Commerce Zipcode Detect URL
+    Default: 'https://ipapi.co/json?key=cKGC3jQbSIoXYmI2KtXObugsKfosD9Yr0HnkHhPUu1SM2wQhE0'
```
BastLeblanc, line 108: is there not any risk of having publicly available keys like this?
Do we have the agreement from these 3rd parties
- ipapi
- fenixcommerce

if not, there should be instructions for the user to request api keys and add them there

aws/cloudformation-templates/base/_template.yaml
```diff
@@ -102,6 +102,52 @@ Parameters:
...
+    NoEcho: true
+
+  FenixEddEndpoint:
+    Type: String
+    Description: Fenix Commerce Estimated Delivery Date Endpoint URL
+    Default: 'https://awsretaildemo.delest.fenixcommerce.com/fenixdelest/api/v2/deliveryestimates'
+
+  FenixMonetaryValue:
+    Type: String
+    Description: Fenix Commerce Default monetary value
```
BastLeblanc, line 123: add a description if it's important for the demo. If we don't expect the customer to change it, have it as a hardcoded parameter to the web-ui service but not the main tpl

aws/cloudformation-templates/template.yaml
```diff
@@ -388,6 +412,51 @@ Parameters:
       Google Analytics Measurement Identifier (optional). If specified, the Google Analytics integration in the web application is activated and
       events are sent to the data stream associated with the Measurement ID.
 
+  FenixZipDetectUrl:
+    Type: String
+    Description: Fenix Commerce Zipcode Detect URL
+    Default: 'https://ipapi.co/json?key=cKGC3jQbSIoXYmI2KtXObugsKfosD9Yr0HnkHhPUu1SM2wQhE0'
```
BastLeblanc, line 418: why is it the default repeated here? a default should be in one place only
+ see my remark on apices being public

aws/cloudformation-templates/web-ui-pipeline.yaml
```diff
@@ -146,6 +146,41 @@ Parameters:
     Type: String
     Description: URL of WebSocket API for Geofence notifications
 
+  FenixZipDetectUrl:
+    Type: String
+    Description: Fenix Commerce Zipcode Detect URL
+    Default: 'https://ipapi.co/json?key=cKGC3jQbSIoXYmI2KtXObugsKfosD9Yr0HnkHhPUu1SM2wQhE0'
```
BastLeblanc, line 152: same. no default if already printed elsewhere

BastLeblanc (changes requested):
see my comments on required changes on the tpl before it is approved.

All the rest looks good to me

aws/cloudformation-templates/base/_template.yaml
ronaks4: Fenix has agreement with ipapi and key used here is has rate limit with quota to prevent over use. this is for demo only key. Future version will have Fenix account creation which will give specific key for account.

ronaks4:
updated description for monetary value

BastLeblanc (approved):
after discussing with Ronak, looks good. 
Default values are repeated in cfn template in case we need to run part of it.

And it's temporary while we have only a demo and not a full fledged workshop

PR #351, opened 2022-04-25
Empty SwaggerUI bucket with a custom resource to not block deletion

ijemmy:
*Issue #, if available:*

*Description of changes:*
When deleting the stack, the `cloudfront.yaml` stack cannot be deleted because the `SwaggerUIBucket` isn't empty.

Example log with error:
```
2022-04-04 10:18:02 UTC+0100SwaggerUIBucketDELETE_FAILEDThe bucket you tried to delete is not empty (Service: Amazon S3; Status Code: 409; Error Code: BucketNotEmpty; Request ID: EPFPPXD4MC1EMGMR; S3 Extended Request ID: lXlm2mDYSA/vrEo34MdNIwsWUn3ax16ZAeFzjDYC/5g3B0eUlN3sl5jZKZcEfMASYw98GB5ANbQ=; Proxy: null)
```

This change add a custom resource to empty the bucket during CloudFormation deletion. 

I have deployed the main stack & deleted it. The deletion doesn't stuck with non-empty bucket now.

### DIFF
--- a/aws/cloudformation-templates/base/cloudfront.yaml
+++ b/aws/cloudformation-templates/base/cloudfront.yaml
@@ -13,13 +13,6 @@ Conditions:
   IADRegion: !Equals [!Ref "AWS::Region", "us-east-1"]
 
 Resources:
-
-  # Empties bucket when stack is deleted
-  EmptyStackBucket:
-    Type: Custom::EmptyStackBucket
-    Properties:
-      ServiceToken: !Ref CleanupBucketLambdaArn
-      BucketName: !Ref WebUIBucket
   
   # Web UI infrastructure    
   WebUIBucket:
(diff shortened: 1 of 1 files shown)


aws/cloudformation-templates/base/cloudfront.yaml
```diff
@@ -29,6 +22,13 @@ Resources:
       WebsiteConfiguration:
         IndexDocument: index.html
 
+  # Empties bucket when stack is deleted
+  EmptyWebUIBucket:
+    Type: Custom::EmptyStackBucket
+    Properties:
+      ServiceToken: !Ref CleanupBucketLambdaArn
+      BucketName: !Ref WebUIBucket
+
```
ijemmy, line 31: Turn out that we have already had this for the `WebUIBucket`. (That is why we never had the issue with WebUI)

I moved this to below the `WebUIBucket` so that it's more logical to read.

james-jory (approved):

PR #358, opened 2022-05-11
Makefile for running integration tests

dairiley:
Issue #, if available:
N/A

Description of changes:
Adds a Makefile with targets to setup and run integration tests, either in bulk or by service. The current integration tests (orders and products) have been altered to load in environment variables using dotenv, and the default values for local testing are added to the .env.template file.


BastLeblanc (approved):
LGTM !

PR #395, opened 2022-09-20
Amplitude recommend personalize workshop

manbearshark:
*Issue #, if available:*

Getting ready to ship a workshop with Amplitude.  This required retail demo store training data to train the Amplitude Recommend Personalize model.

*Description of changes:*

Modified the personalize interactions generator script to be able to generate events for Personalize and Amplitude.  In the process, discovered that the script had a couple issues, one being that the 90 days of data generated is fixed to three months in 2020, the other that the script did not correctly fill events throughout the whole time span specified, rather enede event generation a month early in the generator time span.

The changes made were to create a range 90 days prior to the date / time the script is run, to refactor the event file writer to generate the Personalize interactions file and the Amplitude interactions file simultaneously.  I also added a check to the event generation loop to keep the loop running until the script generates the minimum number of interactions but also to keep going until the script reaches the LAST_TIME timestamp for the interactions.  A more realistic time spacing between interactions was also added:  this_timestamp += random.randint(int(seconds_increment), int(seconds_increment * 2)) rather than a time between 1 second and 11 seconds.

Testing was done to make sure the new script behaves like the old one in the retail demo store deployment. 

### DIFF
--- a/generators/datagenerator/output.py
+++ b/generators/datagenerator/output.py
@@ -5,9 +5,6 @@ from datagenerator.segment import SegmentIdentifyEvent, SegmentTrackEvent, Segme
 from datagenerator.amplitude import AmplitudeIdentifyEvent, AmplitudeTrackEvent, AmplitudeSender
 from datagenerator.file import FileEvent
 
-# TODO: Add Personalize output file formatter
-# TODO: Add Amplitude output formatter
-
 class OutputFormatter:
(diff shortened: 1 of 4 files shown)


manbearshark:
Wanted to get this in the review queue because there are some potentially controversial changes in this - I think these changes will make for a more realistic workshop when using an analytics tool, especially WRT the dates / timestamps.

PR #441, opened 2023-05-22
Pin requests for compatibility

james-jory:
*Description of changes:*

- Pinned requests dependency to 2.29.0 for microservices to avoid conflict with urllib3. https://github.com/boto/botocore/issues/2926

- Triggered build and deployment cycle for microservices in this PR.
- Inspected build logs to make sure there were no dependency warnings/errors.
- Tested functionality of microservices affected by this PR.


gfaires (commented):
There is a new release of requests, 2.31.0, to address security issue: https://github.com/psf/requests/security/advisories/GHSA-j8r2-6x86-q33q. Will this get flagged if we pin to 2.29.0?  
Instead, should we pin to 2.31.0 and add urllib3<2 ?

nateprewitt:
Yes, this was also the recommendation for the Requests 2.30.0 [release notes](https://github.com/psf/requests/blob/main/HISTORY.md#2300-2023-05-03). I would strongly advise against pinning users to older versions of Requests as it's compatible with all versions of Boto3/Botocore.

BastLeblanc (approved):
LGTM

PR #451, opened 2023-06-01
Implement API Gateway

gfaires:
Enables site to be served over https

*Issue #, if available:*

*Description of changes:*
* Move ECS services into private subnet
* Create API Gateway to front services
* Connect APIGW to services through VPC Link
* Lambda Authorizer used for APIGW routes.  Validates tokens if presented, otherwise authorizes all requests
* Moved video service thumbnails to be served from Cloudfront/S3

* Tested OpenSearch, Personalize and Pinpoint workshops
* Full deploy to us-east-1
* Tested running web-ui locally


aws/cloudformation-templates/segment.yaml
```diff
@@ -51,6 +57,10 @@ Resources:
         S3Key: !Sub '${ResourceBucketRelativePath}aws-lambda/segment-personalize-inference-destination.zip'
       Runtime: python3.8
       Timeout: 900
+      VpcConfig:
+        SecurityGroupIds:
+          - !Ref LambdaVpcSecurityGroup
+        SubnetIds: !Split [",", !Ref LambdaVpcSubnets]
```
james-jory, line 63: Since the [Segment Personalize destination](https://segment.com/docs/connections/destinations/catalog/amazon-personalize/) calls this function from Segment's account, we should make sure that this call will still succeed with the function in a VPC.

aws/cloudformation-templates/apigateway.yaml
```diff
@@ -0,0 +1,657 @@
...
+        S3Key: !Sub '${ResourceBucketRelativePath}aws-lambda/apigw-authorizer.zip'
+      Handler: index.handler
+      Runtime: nodejs18.x
+      Role: !GetAtt LambdaAuthorizerRole.Arn
+      MemorySize: 512
+      Timeout: 60
+      Environment:
+        Variables:
+          USER_POOL_ID: !Ref CognitoUserPoolId
+          CLIENT_ID: !Ref CognitoAppClientId
```
james-jory, line 70: I'm sure there's a reason for using a custom authorizer but curious why [API GW Cognito authorizer](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-integrate-with-cognito.html) cannot be used?

james-jory (commented):
Overall the changes look good. Some significant changes here so regression testing will be important. Left a question/comment and flagged the Segment integration function for some closer inspection to make sure it still works.

aws/cloudformation-templates/apigateway.yaml
gfaires: There's a couple of reasons.  First, a Cognito Authorizer is only available for a REST API, whereas we're using a HTTP API.  Second, we need to support _guest_ access, which Cognito User Pools doesn't natively support - therefore, using a Cognito Authorizer would force you into a workaround, such as creating a specific _guest_ user

aws/cloudformation-templates/segment.yaml
gfaires: Good shout.  We can verify by performing a cross account call. 
However, I believe it should be fine since access to the Lambda is controlled by IAM - whereas the changes that have been made to the configuration are to allow the Lambda access to the VPC, which it achieves by deploying an ENI into the subnet.

james-jory (approved):

src/web-ui/src/public/Live.vue
```diff
@@ -208,7 +208,7 @@ export default {
     },
     async loadPlayer () {
       const mediaPlayerScript = document.createElement("script");
-      mediaPlayerScript.src = "https://player.live-video.net/1.0.0/amazon-ivs-player.min.js";
+      mediaPlayerScript.src = "https://player.live-video.net/1.18.0/amazon-ivs-player.min.js";
```
BastLeblanc, line 211: 1.19.0 was released on May 23rd, any reason to not use it ?
Ref: https://docs.aws.amazon.com/ivs/latest/userguide/release-notes.html

BastLeblanc (commented):
Looks good (small Q on IVS versions)

But a general question: How was decided which API routes goes to API gateway, which one don't?

For example:
/users/all : yes agreed as it's not used by the UI, and disclose too much

But all the /products (update, delete), could be useful for other applications integration.
Of course we can add them later, but I wonder if we should just add them to the api gateway integration right now?

src/web-ui/src/public/Live.vue
gfaires: Regarding updating the player. It just means having to update and retest, which I would rather do as part of our next review of all dependencies.

The API GW routes were decided based on what was required by the current frontend.  Service to service communication is direct, as before, so there are no restrictions.  If we wanted to expose sensitive functions to the frontend, then we should update the authorization Lambda to ensure that the user has sufficient privileges - something we don't do now, but should be easy to add now that we have the authorizer lambda in the APIGW.

PR #459, opened 2023-06-29
added cart service functionality in python

Adibuer-lab:
*Issue #, if available:*

*Description of changes:*

In this pull request, I have introduced a Python version of the existing carts Service, which was originally written in Go. By switching to Python, we hope to simplify maintenance and allow more people to contribute effectively.

Here is a summary of the main changes:

Added new Python scripts replicating the functionality of the Product Service, which was originally written in Go. These scripts include src/carts/src/carts-service/app.py, src/carts/src/carts-service/handlers.py, src/carts/src/carts-service/requirements.txt, src/carts/src/carts-service/routes.py, src/carts/src/carts-service/server.py, src/carts/src/carts-service/services.py

Updated the Dockerfile to use a Python 3.10 slim image instead of the Golang image and use python scripts. The Python packages required by the new scripts are installed via the requirements.txt file

Original dockerfile for Go container is now src/carts/Dockerfile-go

Modified src/carts/load_catalog.py to allow setting local dynamodb endpoint as argument and create tables if not created yet

The Dockerfiles and Python scripts have been placed in the same locations as their Go counterparts to keep the structure of the service consistent.

Added persistence to carts service using dynamodb

Added new dynamodb table for carts service

added new GET method to /cart route on api gateway 

created custom origin request policy  and added simple cors response header policy to cdn to allow passing query parameters to service

logs now output to std output for better log analysis

Added unit tests for carts service

deployed locally to test and deployed in aws account to test


aws/cloudformation-templates/services/ProductsCustom.yaml
```diff
@@ -43,6 +46,7 @@ Resources:
                 Resource:
                   - !Sub 'arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${ProductsTable}'
                   - !Sub 'arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${CategoriesTable}'
+                  - !Sub 'arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${CartsTable}'
```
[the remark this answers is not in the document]
Adibuer-lab: You're right it's not necessary. That is left over from when I had to initially load in a cart for the guest user but that was not necessary. Will remove this.

src/carts/src/carts-service/dynamo_setup.py
```diff
@@ -0,0 +1,61 @@
...
+    
+
+    if ddb_endpoint_override:
+        running_local = True
+        app.logger.info("Creating DDB client with endpoint override: " + ddb_endpoint_override)
+        verify_local_ddb_running(ddb_endpoint_override)
+        dynamo_client = boto3.client(
+            'dynamodb',
+            endpoint_url=ddb_endpoint_override,
+            region_name='us-west-2'
```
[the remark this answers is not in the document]
Adibuer-lab: The boto3 client requires a region, either in your aws config environment variables or outlined here like this. In this case since it's the local endpoint it shouldn't really matter what region you put in(it could be any string), but- the local dynamodb has thing where it saves the databases with name, <ACCESS_KEY>_<REGION>. So if you want to query your tables you need to have the right access key and region across your containers. I default to us-west-2 since that is the default region in the env.template and it's easiest to keep track of.

aws/cloudformation-templates/base/cloudfront.yaml
```diff
@@ -98,6 +112,8 @@ Resources:
           TargetOriginId: S3
           ViewerProtocolPolicy: redirect-to-https
           CachePolicyId: !Ref UICachePolicy
+          OriginRequestPolicyId: !Ref UIOriginRequestPolicy
+          ResponseHeadersPolicyId: 60669652-455b-4ae9-85a4-c4c02393f86c
```
dairiley, line 116: Where does this id come from? Can it be parameterized?

src/carts/src/carts-service/dynamo_setup.py
```diff
@@ -0,0 +1,61 @@
...
+# DynamoDB table names passed via environment
+ddb_table_carts = os.getenv("DDB_TABLE_CARTS")
+
+# Allow DDB endpoint to be overridden to support amazon/dynamodb-local
+ddb_endpoint_override = os.getenv("DDB_ENDPOINT_OVERRIDE")
+running_local = False
+
+dynamo_client = None
+
+def verify_local_ddb_running(endpoint):
```
dairiley, line 23: Can you add some docstrings?

src/carts/src/carts-service/dynamo_setup.py
```diff
@@ -0,0 +1,61 @@
...
+    app.logger.info(f"Verifying that local DynamoDB is running at: {endpoint}")
+    for _ in range(5):
+        try:
+            response = requests.get(endpoint)
+            if response.status_code >= 200:
```
dairiley, line 28: This should be more specific so as to only consider response codes within the 200's as any 400 or 500 errors would not be deemed a success

src/carts/src/carts-service/oldservices.py
```diff
@@ -0,0 +1,70 @@
...
+        except BadRequest as e:
+            app.logger.warning(f'Error creating cart: {str(e)}')
+            raise 
+
+    def update_cart(self, cart_id):
+        try:
+            cart = request.get_json(force=True)
+            cart_id_str = str(cart_id)
+            if cart_id_str in self.carts:
+                cart['id'] = cart_id_str
```
dairiley, line 38: This snippet is repetitive, consider moving to it's own method

src/carts/src/carts-service/routes.py
```diff
@@ -0,0 +1,51 @@
...
+def carts_index():
+    app.logger.info('Processing request for /carts')
+    if request.method == 'OPTIONS':
+        return _build_cors_preflight_response()
+    elif request.method == 'POST':
+        response = make_response(jsonify(cart_service.create_cart()))
+        response.status_code = 201
+        response.content_type = 'application/json'
+        return response
+    #return jsonify(cart_service.cart_index()), 200
```
dairiley, line 33: Comment can be removed

src/carts/src/carts-service/services.py
```diff
@@ -0,0 +1,180 @@
...
+from werkzeug.exceptions import BadRequest
+
+
+class CartService:
+    dynamo_client = dynamo_client
+    ddb_table_carts = ddb_table_carts
+    cart_ttl_days = 7
+
+    @staticmethod
+    def marshal_item(data):
```
dairiley, line 23: Just out of curiosity, why do we marshal/unmarshal the cart this way rather than using the json library?

src/carts/test/integ/test_carts.py
```diff
@@ -0,0 +1,41 @@
...
+    endpoint = "/carts?username=:username"
+    params = {":username": test_username}
+    get_request_assert(carts_api_url, endpoint, schemas_path, endpoint_params=params)
+
+def test_get_cart_by_id_should_return_with_correct_schema():
+    endpoint = "/carts/:cart_id"
+    params = {":cart_id": test_cart_id}
+    get_request_assert(carts_api_url, endpoint, schemas_path, params)
+    
+def test_put_cart_by_id_should_return_with_correct_schema():
```
dairiley, line 38: Can you add a couple of tests to validate error handling?

src/products/load_catalog.py
```diff
@@ -57,23 +68,72 @@ def truncate_table(table):
             else:
                 break
     print(f"Deleted {counter} items")
+    
+def  create_table(resource, ddb_table_name, attribute_definitions, key_schema, global_secondary_indexes=None):
```
dairiley, line 72: Extra whitespace after 'def'

src/products/load_catalog.py
```diff
@@ -85,18 +145,48 @@ def truncate_table(table):
...
+                    "KeySchema": [{"AttributeName": "name", "KeyType": "HASH"}],
+                    "Projection": {"ProjectionType": "ALL"},
+                    "ProvisionedThroughput": {
+                        "ReadCapacityUnits": 5,
+                        "WriteCapacityUnits": 5,
+                    }
+                }
+            ]
+        )
+        table = dynamodb.Table(categories_table_name) # type: ignore
```
dairiley, line 188: Remove comment

src/products/load_catalog.py
```diff
@@ -130,4 +255,49 @@ def truncate_table(table):
...
+                        "ReadCapacityUnits": 5,
+                        "WriteCapacityUnits": 5,
+                    }
+                }
+            ]
+                    
+        )
+    enable_ttl_on_table(dynamodb, carts_table_name, ttl_attribute='ttl')
+    
+    """ table = dynamodb.Table(carts_table_name) # type: ignore
```
dairiley, line 289: Can this code block be removed?

aws/cloudformation-templates/base/cloudfront.yaml
Adibuer-lab: This is the id for the [SimpleCORS](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-response-headers-policies.html#managed-response-headers-policies-cors) AWS Managed response header policies for cloudfront. should be able to be parametrized. I'll see if I can make the change, if I can't will ask for help

src/carts/src/carts-service/services.py
Adibuer-lab: I didn't know dynamodb items could be marshalled/unmarshalled using the json library. I'd prefer a standard way if there is one, rather than piecing together my own method. Could you point me to a resource or anything that shows how to do that?

aws/cloudformation-templates/base/cloudfront.yaml
dairiley: No need if it's standard, but I'd probably add a comment so people don't wonder in the future

src/carts/src/carts-service/dynamo_setup.py
dairiley: For most functions more than a couple of lines (unless it's doing something complex in a couple lines). I trust your judgement. The idea is to just provide a simple explanation so contributors can refer to it rather than going through all the code

src/carts/src/carts-service/services.py
dairiley: My bad, not the json library for DynamoDB but boto3 has the TypeSerializer/Deserializer https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/dynamodb/types.html

src/carts/src/carts-service/dynamo_setup.py
```diff
@@ -0,0 +1,61 @@
...
+    else:
+        running_local = False
+        try:
+            response = requests.get("http://169.254.170.2/v2/metadata", timeout=1)
+            metadata = json.loads(response.text)
+            region = metadata['AvailabilityZone'][:-1]
+        except Exception as e:
+            app.logger.warning("Could not fetch ECS metadata, defaulting region to 'us-east-1'")
+            region = 'us-east-1'
+        dynamo_client = boto3.client('dynamodb', region_name=region)
```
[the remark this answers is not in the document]
Adibuer-lab: alright change done now. added AWS_DEFAULT_REGION environment variable in cloudformation template and it uses that now

BastLeblanc (approved):
All remaining issues have been fixed, looks good to merge

PR #491, opened 2023-09-20
adding ACM Cert Arn as a parameter for WebUICDN CloudFront distribution

ccl19:
*Issue #, if available:*

*Description of changes:*
Added the ACM Cert Arn as a parameter, to enable automated blue-green deployment.

Re-staged bucket then redeployed by adding a ACM cert ARN. Distribution is created with the correct certificate attached.

### DIFF
--- a/aws/cloudformation-templates/template.yaml
+++ b/aws/cloudformation-templates/template.yaml
@@ -466,6 +466,10 @@ Parameters:
     Default: 'nr50Qdu7FM94n2X1GYuhA8cFzyvdYTJ5Qka4XMOd'
     NoEcho: true
 
+  ACMCertificateArn:
+    Type: String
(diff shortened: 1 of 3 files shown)


aws/cloudformation-templates/template.yaml
```diff
@@ -466,6 +466,10 @@ Parameters:
     Default: 'nr50Qdu7FM94n2X1GYuhA8cFzyvdYTJ5Qka4XMOd'
     NoEcho: true
 
+  ACMCertificateArn:
```
BastLeblanc, line 469: this parameter should be added in the "Optional Integrations" group to avoid having it floating at the end (it looks like it's part of the Fenix integration)

And add a comment that is optional only required for deployment with a custom CNAME domain

BastLeblanc (approved):
all good now

PR #526, opened 2023-12-08
Fix for integration tests failure

gfaires:
Temporary fix to pass tests.
Removed aliases from required attributes to be returned on Put request.  Issue is caused by JSON being read into a dict, rather than a data object.  Therefore, as the alias isn't passed in on the PUT, it doesn't get returned in the response since there is no data object or additional logic to default the value.

*Issue #, if available:*

*Description of changes:*

Ran integration tests


BastLeblanc (approved):
LGTM

PR #528, opened 2023-12-18
Fix Personalize delete edge cases

### LINKED ISSUE #527: Personalize resource delete corner cases
Ran into some corner cases when deleting the Retail Demo Store stack

- If a Personalize recommender is in an `INACTIVE` state (i.e., stopped), the delete logic will raise an exception. Instead it should just delete the recommender. Although not likely, it is possible for the user to stop recommenders to reduce cost of the deployment.
- The dataset group delete logic optimistically tries to delete the dataset group rather than calling describe to check its status. If the dataset group is already being deleted (such as after a reentrant call following the delete being kicked off), this logic will fail with a ResourceInUseException. This creates a race condition with just this resource types; all the other types have describe calls before attempting deletes.

james-jory:
*Issue #, if available:*

Closes #527 

*Description of changes:*

- Added `INACTIVE` to list of status values that are safe to delete for recommender resource type. Note: this is unique to recommenders since they can be "stopped".
- Added call to describe dataset group resource type before attempting delete in case it's already being deleted.

- Tested changes by deleting Retail Demo Store stack with Personalize resources provisioned.


james-jory:
Replaced by PR #564

PR #545, opened 2024-02-08
Fix for broken frontend

gfaires:
Return price as a float instead of a string.  This changed in recent PR, which broke the frontend

*Issue #, if available:*

*Description of changes:*
Return price as a float instead of a string. Update integration tests schema with price changed from string to number.

Tested web-ui locally and also ran products service integration tests


BastLeblanc (approved):
Looks good. Thanks

PR #547, opened 2024-02-16
Bump python3.7 to python3.12

james-jory:
*Description of changes:*

Support for python 3.7 was recently dropped from AWS Lambda. This was causing deployments to fail for 3 functions that still used python3.7 as the runtime. This PR updates the runtime to the latest supported version, python3.12. 

There are other functions that use the 3.8, 3.9, 3.10, and 3.11 runtimes that also need to be upgraded but those aren't addressed in this PR since they are supported and to keep the surface area of this urgent PR small.

Tested a full deployment and delete deployment. Succeeded without errors.


gfaires (approved):
LGTM

PR #564, opened 2024-03-27
Personalize pre-create/delete resource improvements

### LINKED ISSUE #527: Personalize resource delete corner cases

james-jory:
*Issue #, if available:*

Closes #527 

*Description of changes:*

- Added support for the Personalize content generator to the pre-create resource logic. This required moving the function into the VPC so that it had access to the Products service to decorate featured products with theme info.
- Added logic to verify that existing schema design matches expected schema to better handle edge case of pre-existing schema that may not match.
- Only download/upload CSVs as needed rather than on every function invocation.
- Improved robustness of the Personalize resource delete logic
    - Added support for all status transitions for async resource types to better handle cases where resources are in various transition status (i.e., being created) that prevent them from being deleted.
    - Added support for stopping solution versions (training) that are currently being created so they can be cleanly deleted.
    - Added logic to wait for DSG to be deleted
- Added logic to the Personalize content generator lab/workshop to handle regression in feature of output file naming

- Tested the delete resource script multiple times when resources were in different statuses.
- Tested full deployment and delete process of the project


BastLeblanc (approved):
Great job on this PR! approved

PR #631, opened 2024-09-05
Copy external product images from source to target bucket

dpsaws:
*Issue #, if available:*

*Description of changes:* The custom product images uploaded from datasets/1.4/images into the source bucket needs to be copied to the webui destination bucket.


BastLeblanc (approved):
Good to me, thanks for the contribution

Issue #668, opened 2024-12-01
UnboundLocalError from `index.handler` during deployment


MustaphaU:
When deploying the Retail Demo Store, the following error is encountered when **retries are exhausted** during the execution of the `handler` function in the `LoadDataLambdaFunction` in [deployment-support.yaml](https://github.com/aws-samples/retail-demo-store/blob/master/aws/cloudformation-templates/deployment-support.yaml#L163-L217).

Error
 ```bash
  UnboundLocalError: cannot access local variable 'e' where it is not associated with a value
  Traceback (most recent call last):
    File "/var/task/index.py", line 31, in handler
      logger.error(f"Error calling product service init: {e.code} : {e.reason}")
```

The relevant code block: 

```python
...
while retries < max_retries:
      try:
          with urlopen(request) as response:
              logger.info(f"Product Service init method success: {response.read()}")
              # exit while, success.
              break
      except URLError as e:
          retries += 1
          print(f"Request failed. Retrying in 30 seconds... (Attempt {retries}/{max_retries})")
          print(f"Error: {e}")
          time.sleep(30)

#
if retries >= max_retries:
    logger.error(f"Error calling product service init: {e.code} : {e.reason}")
    response_status = cfnresponse.FAILED
    response_data['Message'] = f"Resource {event['RequestType']} failed: {e}"  
...      
```
The scope of `e` is local to the `except` block and Python does not persist it outside that scope. The condition `if retries >= max_retries` runs outside the loop and, therefore, cannot access `e`. The value (`e`) within the block can be assigned to a variable to be accessed outside.

BastLeblanc:
Thanks for the report. 
The underline issue on the product service is fixed by #670 

Happy to accept PR for the fix on the lambda python code of the LoadDataLambdaFunction, adding this on the backlog

PR #669, opened 2024-12-01
Update deployment-support.yaml to fix `UnboundLocalError`

MustaphaU:
*Issue #, if available:*
#668 

*Description of changes:*
This PR updates the `index.handler` function for the `LoadDataLambdaFunction` to ensure no undefined variables are referenced. A new variable, `error`, is introduced to track the most recent exception during retries. The `logger.error` statement has been updated to reference `error` instead of `e`, ensuring the function handles errors without encountering an UnboundLocalError. Also, since `URLError` has no `code` attribute, only the `reason` attribute is used.

*relevant section with updates*
```python
while retries < max_retries:
      try:
          with urlopen(request) as response:
              logger.info(f"Product Service init method success: {response.read()}")
              # exit while, success.
              break
      except URLError as e:
          error = e  #e is assigned to a variable
          retries += 1
          print(f"Request failed. Retrying in 30 seconds... (Attempt {retries}/{max_retries})")
          print(f"Error: {e}")
          time.sleep(30)

#
if retries >= max_retries:
    logger.error(f"Error calling product service init: {error.reason}") 
    response_status = cfnresponse.FAILED
    response_data['Message'] = f"Resource {event['RequestType']} failed: {error}" 
```

1. Deployed the CloudFormation template with the updated Lambda function.
2. Confirmed that exceptions are logged correctly without triggering an `UnboundLocalError`


BastLeblanc (approved):
LGTM thanks for the fix

